PR #11, opened 2026-04-29
extract: skip Rust functions annotated with #[test]

jschwe:
Walk backward through the contiguous run of attributes, blank lines, and comments immediately preceding each `fn` keyword. If `#[test]` is found in that run, skip the function and advance past its closing brace.


yuesong-feng:
I think it's better to keep it simple. #12 can cover 99% LLM-generated rust test cases.

Issue #7, opened 2026-04-25
Support resuming a session


jschwe:
If a session gets interrupted, e.g. because of not enough balance, it would be great if we could resume the session after topping up and don't have to start from scratch. 
If this works only when running locally, that would be fine.

haoran-ding:
This is really a good suggestion. Maybe it can be realized by asking LLM to check the current progress before doing anything. We will attempt to figure out whether it is feasible.

haoran-ding:
We have implemented the feature internally. After checking it carefully, we will release it.

PR #21, opened 2026-06-03
feat(pipeline): add --resume to continue an interrupted run

heiyan-2020:
## Summary

- `run_pipeline` wiped `fm_agent/` unconditionally at startup, so every invocation restarted from scratch — an interrupted run lost all prior specs and verification results.
- Adds a `--resume` flag (or `FM_AGENT_RESUME=1`) that continues a previous run: keeps existing progress and only does the remaining work.
- Fresh-run behaviour is unchanged.

## Changes

- **feat(pipeline): add --resume to continue an interrupted run**
  - resume keeps `fm_agent/` instead of `rmtree` (preserves `phases.json`, generated specs, logic-verification + bug-validation results)
  - skips the `setup_context` LLM call when `phases.json` already exists (reuses the phase plan)
  - extracts with `force=False` so already-specced files keep their specs (equivalent to `force=True` on a fresh run, where `fm_agent/` is empty)
  - adds a skip-if-exists guard to `_validate_single_bug` so bug validation isn't paid for twice on resume

## Diff Stats

```text
 main.py             | 38 +++++++++++++++++++++++++++++++-------
 src/verification.py |  9 +++++++++
 2 files changed, 40 insertions(+), 7 deletions(-)
```

---
Generated by an AI coding agent


haoran-ding:
Please also update README

haoran-ding:
Besides, the PR only considers resuming spec generation and bug validation. It does not consider code reasoning.

nhaorand:
Fix issue #7

Issue #14, opened 2026-05-07
support language server


gucliti:
Enhance language server plugins with support for specified programming languages, ensuring the reliable and straightforward generation of function call chains.

haoran-ding:
We find another tool called codegraph (https://github.com/colbymchenry/codegraph), which seems to be also useful. We have supported most languages with it (PR #63). For other languages, we will continue to find useful tools.

haoran-ding:
We plan to support erlang with https://github.com/whatsapp/erlang-language-platform

PR #22, opened 2026-06-04
Add optional LSP analysis support and pipeline stage controls

n3sRan:
## Summary

Add optional LSP analysis support and pipeline stage controls. This PR introduces an extensible LSP framework, initial Go support through `gopls`, and CLI options for running partial or LSP-only pipeline flows.

## Changes

- Add `--stop-after` to stop the pipeline after a selected stage.
- Add `--enable-lsp` to explicitly enable LSP analysis.
- Add `--lsp-only` to run LSP analysis and exit before codebase understanding.
- Add a provider-based LSP module with registry, client, models, runner, and initial Go provider.
- Generate LSP artifacts under `fm_agent/lsp/`, including `status.json`, `symbols.json`, and `calls.json`.
- Use LSP symbols during function extraction when available.
- Use LSP-derived call information during top-down layer generation when available.
- Keep LSP disabled by default and allow graceful fallback unless strict mode is enabled.
- Update English and Chinese README documentation for the new options and LSP workflow.

## Testing

- Verified CLI help output for --stop-after, --enable-lsp, and --lsp-only.
- Verified compile checks for the updated Python modules.
- Tested LSP analysis on a lightweight Go repository with gopls, confirming symbol and call artifacts are generated successfully.


haoran-ding:
This PR resolves issue #14

haoran-ding:
We convert to use codegraph for most languages and re-org the code to be more modular. So this PR will be closed. Still appreciate the contribution,

PR #53, opened 2026-06-25
feat(extract): add codegraph backend for Python

Dshuishui:
## Summary
Integrate `@colbymchenry/codegraph` (tree-sitter based) as the function extraction and call graph backend for Python, replacing the regex-based approach. Existing regex extractor is preserved as a fallback when codegraph is unavailable.

## Changes
- `src/extractors/codegraph.py` — new module: queries `.codegraph/codegraph.db` for function bodies and call edges
- `src/extract.py` — route Python files through codegraph when index is available; fall back to regex otherwise
- `src/generate_topdown_layers.py` — build call graph from codegraph edges when available; fall back to regex scan otherwise
- `main.py` — auto-run `codegraph init` between Stage 1 and Stage 2 so only original source files are indexed (before `fm_agent/extracted_functions/` is created)
- `install.sh` — install `@colbymchenry/codegraph` via bun as part of FM-Agent setup

## Motivation
The regex-based extractor is fragile on complex syntax. Tree-sitter based codegraph provides accurate AST-level function extraction and precise call edges. The new backend is designed to be extensible: adding support for a new language requires only adding one line to `CODEGRAPH_SUPPORTED` in `codegraph.py`.

As a side effect, the codegraph path also correctly extracts `async def` functions, which the regex-based fallback silently drops (see #54).

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -82,5 +82,13 @@ fi
 echo "[..] installing/updating oh-my-openagent"
 bunx oh-my-openagent install --no-tui --claude=no --gemini=no --copilot=no
 
+# ---------- codegraph ----------
+if command -v codegraph &>/dev/null; then
+    echo "[ok] codegraph found: $(codegraph --version 2>/dev/null || echo 'unknown version')"
+else
+    echo "[..] installing codegraph"
+    bun install -g @colbymchenry/codegraph
+fi
+
 echo ""
 echo "=== all dependencies installed ==="
--- a/main.py
+++ b/main.py
@@ -16,6 +16,7 @@ from src.opencode_trace import (
     start_opencode_traced,
 )
 from src.incremental_reasoner import run_incremental_pipeline
+from src.languages.codegraph import try_codegraph_init
 import os
 import sys
 import argparse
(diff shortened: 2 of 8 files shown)


haoran-ding:
Please fix the following problems:

- The [try-exception block](https://github.com/fmagent-project/FM-Agent/pull/53/changes#diff-450848dbd32d25cb640a4349c76128577d6dde9707c3f807c283cbb137acc80aR335 ) seems useless 
- All codegraph functions should be defined in codegraph python file rather than main.py. main.py only invokes them.
- Rename the extract folder to be language

Dshuishui:
Thanks for the detailed feedback! Here's what was addressed in the latest commit (c11f6d2):

1. **Removed the useless try-except block** — The `codegraph --version` pre-check was redundant. Replaced with a single `subprocess.run(["codegraph", "init"], ...)` call inside a `try/except FileNotFoundError`, which handles the "not installed" case directly without the extra round-trip.

2. **Moved all codegraph logic into `src/languages/codegraph.py`** — `_try_codegraph_init` (now renamed `try_codegraph_init`) has been moved from `main.py` into the codegraph module. `main.py` now only imports and calls it.

3. **Renamed `src/extractors/` → `src/languages/`** — Updated all import paths in `main.py`, `src/extract.py`, and `src/generate_topdown_layers.py` accordingly.

PR #78, opened 2026-07-03
Add pre-flight environment checks before pipeline run (fixes #25)Add pre-flight environment checks before pipeline run

### LINKED ISSUE #25: Add pre-check to ensure environment is suitable
From the readme:

> 
> FM-Agent uses oh-my-openagent plugin to enhance OpenCode. The comment-checker hook built into this plugin should be disabled, otherwise it may intercept every comment block that FM-Agent writes, which are specifications of functions. It may force the agent to waste tokens justifying or removing them. You can open your oh-my-openagent config file (typically ~/.config/opencode/oh-my-openagent.json) and add disabled_hooks:
> 
> {
>   "disabled_hooks": ["comment-checker"],
> }

It would be nice if FM-Agent could do a quick check in advance and inform the user of the missing configuration and potential issues, if the environment is not setup suitably. The result could be saved as a memory, so the user is not asked again if they wish to ignore the suggestion.

ji-ji-hao:
Add env_check module that verifies the runtime setup is suitable before launching the pipeline:

- Checks LLM_API_KEY is configured in .env
- Checks oh-my-openagent is installed
- Checks comment-checker hook is disabled (FM-Agent writes function specs as comment blocks, which the comment-checker intercepts)

Users can choose to proceed, permanently ignore the warnings, or quit to fix the issues. Non-interactive sessions (CI) auto-proceed with warnings logged.

Closes #25

### DIFF
--- a/main.py
+++ b/main.py
@@ -572,11 +572,6 @@ def frozen_worktree(proj_dir, exclude=("fm_agent",), copy_excluded=True):
 
 
 def run_pipeline(proj_dir, resume=False, required_source_files=None):
-    # ---- pre-flight environment check ----
-    import config
-    from src.env_check import run as env_check_run
-    if not env_check_run(proj_dir, config):
-        sys.exit(0)
     if not os.path.isdir(proj_dir):
         print(f"[Pipeline] ERROR: proj_dir does not exist or is not a directory: {proj_dir}")
         sys.exit(1)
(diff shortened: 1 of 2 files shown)


haoran-ding:
The position of running env_check_run is incorrect. Check happens only when run_pipeline is executed. But the main.py may also run other functions, such as incremental reasoning. Please change the check position.

haoran-ding (approved):
Approved

PR #80, opened 2026-07-04
Feat/erlang elp backend

### LINKED ISSUE #14: support language server

gesi519:
Closes #14
## Summary

  This PR adds Erlang function extraction and direct call-graph support using the
  Erlang Language Platform (ELP) over LSP.

  ## Changes

  - Register Erlang as a supported language.
  - Extract Erlang functions using `textDocument/documentSymbol`.
  - Extract direct call edges using the LSP call hierarchy API.
  - Open all Erlang source files before analysis to support multi-file projects.
  - Reuse the original source-project analysis when generating topdown layers.
  - Persist ELP analysis results to `.codegraph/erlang_callgraph.json`.
  - Implement `function_spans` using ELP source ranges.
  - Add Erlang support to parsing, prompts, verification, test-file detection,
  documentation, and environment configuration.
  - Cache source text and line offsets to reduce repeated file I/O and range conversion
  work.

## Known Limitation

  The existing bug-validation workflow may time out when multiple Erlang probe agents
  run concurrently and share build artifacts. This issue is independent of ELP
  extraction and call-graph generation and should be handled separately.


haoran-ding:
Please also change install.sh to automatically install necessary software and specify version requirements in readme

gesi519:
```bash
  ./install.sh --with-erlang

  This option installs or verifies:

  - Erlang/OTP 26+
  - rebar3 3.24.0+
  - a compatible ELP release

  The default ./install.sh behavior remains unchanged, so users analyzing other languages do not need to install the
  Erlang toolchain.

  The English and Chinese README files now document the version requirements, installation command, platform-specific
  behavior, and the fact that the Erlang integration has been tested on Ubuntu. On macOS, the script uses the current
  Homebrew formula versions; the macOS Erlang setup has not been tested as part of this change.

gesi519:
The timeout issue has been largely resolved. It is recommended to avoid using project directory paths that contain `test` or `tests`

PR #81, opened 2026-07-05
fix: crash on C++ operator/ function name

### LINKED ISSUE #66: C++ operator-overload function names (operator/) crash extraction
## Summary

`run_extraction` writes each extracted function to `<func_name>.<ext>`. C++
operator overloads are extracted (via codegraph) under their literal names, e.g.
`operator/` for a division overload. The `/` is a path separator, so the write
targets a non-existent subdirectory and raises `FileNotFoundError`, aborting the
whole pipeline. Any C++ project with a `operator/` / `operator/=` overload
crashes.

## Reproduction

`frac.cpp`:

```cpp
struct Frac {
    int num, den;
    Frac operator/(const Frac &rhs) const { return Frac{num * rhs.den, den * rhs.num}; }
    int value() const { return num / den; }
};
```

Run the pipeline (Stage 1 builds the codegraph index; Stage 2 extraction crashes):

```
[Pipeline] Stage 1/4: ... codegraph index built.
[Pipeline] Extracting functions from source files...
Traceback (most recent call last):
  File ".../src/extract.py", line 730, in run_extraction
FileNotFoundError: [Errno 2] No such file or directory:
  '.../fm_agent/extracted_functions/frac-cpp/operator/.cpp'
```

Nothing is extracted — the run aborts on the first operator-overload name.
First hit on real code: DuckDB's `src/common/types/hugeint.cpp` defines
`hugeint_t::operator/`.

## Root cause

`src/extract.py::run_extraction`:

```python
os.makedirs(out_dir, exist_ok=True)                      # creates out_dir only
out_file = os.path.join(out_dir, f"{func_name}.{ext}")   # func_name = "operator/"
with open(out_file, 'w') as f:                           # out_dir/operator/.cpp -> ENOENT
```

The `/` in `operator/` is treated as a path separator; the intermediate
`operator/` directory is never created.

## Notes

- codegraph-path specific: the regex extractor's name pattern needs an identifier
  immediately followed by `(`, which `operator/(` breaks, so it never yields
  `operator/`.
- On macOS/Linux only `/` is fatal (path separator); other operator names
  (`operator<<`, `operator*`, `operator[]`, ...) are legal Unix filenames but
  would be illegal on Windows.

## Fix direction

Sanitize path-unsafe characters in function names when forming the output
filename (at minimum `/`), keeping extraction / FQN / call-edge naming
consistent.

tanbing117:
Introduce canonicalize() to sanitize filesystem-unsafe characters in function names (e.g. operator/ -> operator_). Applied at all five data entry points in codegraph and extract layers. Downstream modules need zero changes - FQNs and scope lookups automatically pick up the canonicalized names.

Closes #66

### DIFF
--- a/src/extract.py
+++ b/src/extract.py
@@ -193,6 +193,15 @@ def _strip_angle_brackets(text):
 def _extract_func_name_brace(signature_text, lang_cfg):
     """Extract the function name from a brace-delimited language signature."""
     lang_keywords = lang_cfg["keywords"]
+
+    m = re.search(
+        r'\b(operator\s*(?:\[\]|\(\)|[+\-*/%&|^~!=<>]+|new(?:\s*\[\s*\])?|delete(?:\s*\[\s*\])?))'
+        r'\s*\(',
+        signature_text,
+    )
+    if m:
+        return m.group(1)
+
     cleaned = _strip_angle_brackets(signature_text)
     for m in re.finditer(r'\b(\w+)\s*\(', cleaned):
         name = m.group(1)
--- a/src/extract.py
+++ b/src/extract.py
@@ -6,6 +6,7 @@ import shutil
 import logging
 
 from src.file_utils import is_file_ready, _is_test_file
+from src.languages.codegraph import canonicalize
 from src.languages.registry import batch_extract_all, function_spans_for_file
 
 LANG_CONFIG = {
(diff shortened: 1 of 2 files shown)


haoran-ding:
Please resolve the conflict

haoran-ding (changes requested):
I test the code with the attached file in the issue. It cannot identify operator "/".  Please check why and fix it. If it is the bug of codegraph, please submit an issue to codegraph repo and report it here as well.

Dshuishui:
I verified the fix against the issue #66 repro (`frac.cpp`) and ran the full pipeline end-to-end — extraction no longer crashes and `operator/` is correctly identified and sanitized to `operator_`. Two things worth considering:

**1. Filename safety beyond `/`.** `canonicalize()` currently only replaces `/`. The issue used `/` just as the example because it's the only char that's *fatal* on macOS/Linux (path separator), but it isn't the only unsafe one — other overloads produce names like `operator<<`, `operator>>`, `operator|`, `operator*`, `operator[]`, `operator<`, `operator>`. Those are legal on Unix (so they don't crash there), but `< > : " | ? *` are illegal filename characters on Windows, so those overloads would still break the pipeline on Windows. Might be worth sanitizing the full set of path-unsafe characters rather than only `/`.

**2. Call edges into operator overloads.** After the fix, a caller that invokes an overload (e.g. `a / b`) still gets no call edge to `operator/` — I verified `divide -> operator/` is missing from the call graph. This turned out to be a codegraph-side limitation (codegraph doesn't record call edges into operator-overload methods, even for explicit `a.operator+(b)` syntax); I've filed a related issue upstream at colbymchenry/codegraph#1247.

haoran-ding (approved):
The issue has been resolved by the PR. There are some bugs of codegraph rather than our code.

Issue #68, opened 2026-07-01
Feat: Add an interface for users to provide domain knowledge


haoran-ding:
Add an interface for FM-Agent. The interface allows users to provide domain knowledge with extra markdown files.

haoran-ding:
Fixed by #91

PR #84, opened 2026-07-06
feat: complete the issue #68 and add a feat for knowledge domain

Logic-1729:


haoran-ding:
There are many files that should not be pushed, such as xxx.identifier. Besides, the PR incorrectly revert the code of supporting erlang.

heiyan-2020:
@codex review

haoran-ding:
The method of resolving conflict directly removes some useful code, such as supporting erlang. The PR still cannot be merged.

PR #85, opened 2026-07-06
fix: bound spec generation concurrency

### LINKED ISSUE #83: spec_generation spawns unbounded concurrent opencode processes causes OOM
main.py's spec-generation loop calls start_opencode_traced (non-blocking) for every pending batch of the layer at once. Concurrency = batch count, which scales with the target project.

On duckdb project, one layer produced ~93 batches → 93 concurrent opencode (bun) processes, this may cause OOM or LLM backend overloaded.

All other stages are bounded (ThreadPoolExecutor(max_workers=MAX_WORKERS)); this is the only unbounded path.

Possible fix: cap live spec agents at MAX_WORKERS (chunk the spawn loop, or put a semaphore inside start_opencode_traced/finish_opencode_trace).

okcomputer2000:
Fixes #83

## What changed
- cap live spec-generation agents at `MAX_WORKERS`
- process pending spec batches in chunks instead of spawning one opencode process per batch
- skip empty resumed chunks without invoking `streaming_reasoner`

## Why
`spec_generation` was the only unbounded concurrent path in the pipeline. Large layers could spawn dozens of opencode processes at once, which could cause OOM or overload the LLM backend.

## Validation
- `python3 -m py_compile main.py`

### DIFF
--- a/main.py
+++ b/main.py
@@ -1,4 +1,5 @@
 from config import (
+    MAX_WORKERS,
     OPENCODE_MAX_RETRIES,
     OPENCODE_SPEC_MODEL,
     OPENCODE_MODEL_PROVIDER,
@@ -16,9 +17,8 @@ from src.verification import streaming_reasoner
 from src.extract import run_extraction, EXT_TO_LANG
 from src.generate_topdown_layers import generate_topdown_layers
 from src.opencode_trace import (
-    finish_opencode_trace,
     function_id_from_extracted_path,
-    start_opencode_traced,
+    run_opencode_traced,
 )
 from src.cli_backend import build_agent_command, is_cli_backend_enabled
 from src.incremental_reasoner import run_incremental_pipeline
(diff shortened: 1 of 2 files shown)


heiyan-2020:
@codex review

heiyan-2020:
This chunking strategy, which partitions work based on max_workers, may result in an actual level of concurrency lower than the configured max_workers. 

For example, if a chunk contains a straggler, the entire chunk is blocked, preventing the next iteration from starting until the straggler finishes.

okcomputer2000:
Updated. I replaced the chunk-based scheduling with a rolling worker pool backed by `ThreadPoolExecutor(max_workers=MAX_WORKERS)`, so completed slots can immediately pick up the next pending batch. I also updated `streaming_reasoner` to handle future-backed spec tasks.

heiyan-2020:
looks good to me

PR #90, opened 2026-07-08
Support supplemental call graph edges

### LINKED ISSUE #89: Add feature: Support supplemental call graph edges
Add a new argument "--extra-edge" to main.py, which specifies a JSON file or directory containing supplemental caller-to-callee edges. Then, FM-Agent will include these edges when building call graphs for full, entry-scoped, and incremental reasoning mode.

lar0129:
Closes #89

## Summary
- add `--extra-edge` support for loading supplemental caller-to-callee edges from JSON files or directories
- apply extra edges across full, entry-scoped, and incremental reasoning mode
- use callee aliases when matching caller's `[INFO]` expectation

## Testing
- Tested `--extra-edge` with `docs/extra-edge/example.json`; supplemental edges were applied successfully


docs/extra-edge/example.json
```diff
@@ -0,0 +1,36 @@
+{
+  "edges": [
+    {
+      "caller": {
+        "fqn": "third_party::musl::porting::liteos_a::user::src::exit::exit-c::exit",
+        "callsite_names": [
```
heiyan-2020, line 6: what's the difference between caller's fqn and callsite_name?

lar0129: You can refer to the README_zh.md: 
- callsite_name means “for any function whose body contains a call expression with this name, add an edge from that function to callee.fqn".

lar0129: caller.fqn and caller.callsite_names are two caller-discovery modes. 
- If we can identify the exact caller function from source code, we use caller.fqn, which only adds the edge from that specific function. 
- If we cannot resolve the caller FQN but we know the source callsite name, we use caller.callsite_names to find functions that contain that callsite and add the edge from them.

Dshuishui:
Please resolve the merge conflict first — the branch is currently `CONFLICTING` against `main`.

I then looked through the diff, and the two issues Codex flagged both look real. The edge gets added to the call graph fine, but the caller expectation never reaches the callee's spec:

**1. Exact-only edges (`caller.fqn`, no `callsite_names`):** `build_prompt` only tells a function about its callers, never its callees, so the supplemental callee is never put in the caller's prompt. Since the caller only writes `[INFO]` for calls visible in its source — and an exact edge is exactly the invisible case — no `[INFO]` is written, and the callee gets no caller expectation.

**2. Cross-phase edges:** each phase's graph is built separately, so an edge across phases is dropped — while processing the caller's phase the callee isn't in `phase_fqns` (it survives only in `all_callees`, which the prompt ignores), and in the callee's phase the caller is never scanned. Result: the callee's `phaseN_callers` and aliases are empty. Note `docs/extra-edge/example.json` is itself cross-phase (musl → kernel), so the main example hits this.

Root cause is shared: expectations only flow caller `[INFO]` → callee, and the caller only writes that `[INFO]` for source-visible calls. Fix needs to (a) inject supplemental callees into the caller's prompt, and (b) carry the relationship across phases.

haoran-ding (changes requested):
Please resolve the conflict. Table of Contents in readme should not be changed.

lar0129:
> Please resolve the merge conflict first — the branch is currently `CONFLICTING` against `main`.
> 
> I then looked through the diff, and the two issues Codex flagged both look real. The edge gets added to the call graph fine, but the caller expectation never reaches the callee's spec:
> 
> **1. Exact-only edges (`caller.fqn`, no `callsite_names`):** `build_prompt` only tells a function about its callers, never its callees, so the supplemental callee is never put in the caller's prompt. Since the caller only writes `[INFO]` for calls visible in its source — and an exact edge is exactly the invisible case — no `[INFO]` is written, and the callee gets no caller expectation.
> 
> **2. Cross-phase edges:** each phase's graph is built separately, so an edge across phases is dropped — while processing the caller's phase the callee isn't in `phase_fqns` (it survives only in `all_callees`, which the prompt ignores), and in the callee's phase the caller is never scanned. Result: the callee's `phaseN_callers` and aliases are empty. Note `docs/extra-edge/example.json` is itself cross-phase (musl → kernel), so the main example hits this.
> 
> Root cause is shared: expectations only flow caller `[INFO]` → callee, and the caller only writes that `[INFO]` for source-visible calls. Fix needs to (a) inject supplemental callees into the caller's prompt, and (b) carry the relationship across phases.

Thanks for the detailed review. 
1. The merge conflict against the latest main has been resolved. 
2. For exact-only edges, the intended behavior is to use info_names aliases during callee spec generation to retrieve the caller’s existing [INFO]; the edge is not intended to force the caller to synthesize new [INFO]. 
3. Cross-phase edges are intentionally unsupported, as phase isolation is part of the current design.Therefore, I haven’t changed either behavior in this PR.

haoran-ding (approved):
Thanks

PR #93, opened 2026-07-09
Fix function signature used as filename causing OSError (fixes #82)

### LINKED ISSUE #82: Entry-mode extraction fails on DuckDB with `OSError: [Errno 36] File name too long`
FM-Agent entry mode fails during extraction on DuckDB because an extracted C/C++ function name is
used directly as the output filename.

Command:

  UV_CACHE_DIR=/home/zz/project/FM-Agent/.uv-cache \
  /home/zz/snap/code/247/.local/bin/uv run python main.py /home/zz/project/duckdb \
    --entry-func test::sql::index::art::issue_6799_entrypoint-
    cpp::Issue6799SearchEqualJoinNoFetchMissRegression \
    --end-func src::execution::index::art::art-cpp::SearchEqualJoinNoFetch

 ## Error

  OSError: [Errno 36] File name too long:
  .../extracted_functions/extension/json/yyjson/include/yyjson-hpp/yyjson_mut_val
  *yyjson_mut_obj_with_str(yyjson_mut_doc *doc,
                                                            const char **keys,
                                                            const char **vals,
                                                            size_t count).hpp

  Trace ends at:

  src/extract.py:739

  ## Cause

  src/extract.py uses the extracted function name directly as a filename:

  out_file = os.path.join(out_dir, f"{func_name}.{ext}")

  For some C/C++ headers, the extracted function name can be a long multiline declaration, exceeding
  filesystem filename limits.

  ## Expected

  Extraction should use a safe filename mapping, e.g. sanitized/truncated basename plus stable hash,
  instead of raw func_name.

ji-ji-hao:
Fixes #82.

Tree-sitter sometimes stores full function signatures in its `name` column for macro-generated C/C++ declarations (e.g. DuckDB's yyjson), which were used directly as filenames and exceeded filesystem limits.

Two-layer fix:

- `src/languages/codegraph.py` — new `_bare_function_name()` extracts the bare identifier from signature strings in all three codegraph name-outflow paths
- `src/extract.py` — new `_safe_filename()` replaces `/`, truncates names >200 chars with a stable hash suffix, and falls back to `_function` for empty names

### DIFF
--- a/src/extract.py
+++ b/src/extract.py
@@ -5,9 +5,8 @@ import sys
 import shutil
 import logging
 
-from src.file_utils import is_file_ready, _is_test_file
-from src.languages.codegraph import canonicalize
-from src.languages.registry import batch_extract_all, function_spans_for_file
+from src.file_utils import is_file_ready
+from src.languages.registry import batch_extract_all
 
 LANG_CONFIG = {
     "cpp": {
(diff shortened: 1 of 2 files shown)


haoran-ding (commented):
_bare_function_name looks a little ad-hoc. Does it work for all languages supported by FM-Agent now?

haoran-ding (changes requested):
Please resolve the conflict. Anther merged PR #81 has replaced "/" in file name.

ji-ji-hao:
@haoran-ding Thanks for the review. All feedback addressed:

Rebased onto latest main to incorporate PR #81's canonicalize() and FQN-based call edges
_bare_function_name() runs before canonicalize() at all three codegraph entry points (_node_fqn_map, get_functions_by_file, get_function_spans)
_safe_filename() handles empty-name fallback (_function) at the extraction layer
Only two files changed: codegraph.py and extract.py

Please take another look when you can.

haoran-ding (approved):
The PR resolves the issue #82 successfully.

PR #94, opened 2026-07-09
Harden Verilog/Chisel HDL spec generation: resume trust chain, per-callee gate, LLM-JSON shape safety

Buyi-wsgzg:
## Summary

Ten commits hardening the Verilog/SystemVerilog + Chisel spec-generation flows added in #86, driven by multiple rounds of independent review plus real end-to-end validation (gpt-5.5 and DeepSeek, riscv-mini Chisel/Verilog and hand-written SERV RTL).

### Resume trust chain
- **Stale derived state on rerun-setup resume**: when a reused `groups.json` can't be resumed, derived artifacts (domain context, phases.json, file list, topdown layers, batch prompts, and the rejected `groups.json` itself) are reset before setup reruns — previously a stale manifest could steer Stage 1 into a "continue where you left off" prompt for a manifest this flow already rejected.
- **Partial-missing sources**: the zero-modules guard's fallback now triggers on *any* missing listed source under a reused manifest, not only when every source is gone — a manifest with one surviving and one renamed/deleted file used to spec the survivor and silently never spec the renamed one.
- **Language normalization**: `_groups_to_phases` forces `languages=["chisel"]` unconditionally (mirrors the existing Verilog `_force_verilog_phase_languages`) — the resume veto accepts `"scala"` as a plausible LLM spelling, but the batch-prompt language router only recognized the literal `"chisel"` token, so an accepted `"scala"` manifest silently misrouted to software-style prompts.
- **Hardware-language guard**: the default pipeline's guard now harvests hardware-language tokens from any JSON shape (string/list/dict) with an unambiguous-extension/source-suffix backstop when no token is found, closing a fail-open gap for declared-but-unintelligible language shapes.

### Verilog per-callee submodule gate
- Verilog's instantiation graph is precise (verible CST), so non-leaf modules now must document a `# Submodule: <name>` entry for *every* instantiated child, not just claim `(no submodules)` or contain any single heading. Missing entries are deleted and fed back to the retry prompt with the full required set (avoiding ping-pong on partial fixes). Chisel deliberately keeps its existing advisory-only gate — real-run data shows its graph gaps are ~100% false edges (companion objects, IO bundles, constant objects), so blocking there would strand runs on phantom documentation demands.
- A verible-drift canary detects when the installed verible version's instantiation CST tags are no longer recognized (partial drift beyond a full parse failure) and falls back to the regex scanner instead of silently under-reporting edges.
- `FM_AGENT_NO_VERIBLE=1` (regex-fallback) mode keeps the gate blocking (the fallback only ever misses edges, never invents ones — unlike Chisel's problem) but prints a one-time warning naming the narrower guarantee.

### LLM-JSON shape safety
`groups.json`/`phases.json` are LLM-written, so a single-element field is plausibly emitted as a bare string instead of a one-element list, and a single dependency as a bare int/string. Normalized at the point each fact is established (not re-derived per consumer) across `source_files`, `depends_on_phases`/`depends_on_subsystems`, `languages`, and `file_extensions` — closing several silent mis-parses (character-by-character iteration, phase-id lookup misses, `build_ext_to_lang` routing failures).

### Declared-name single source of truth
`generate_topdown_layers` now stamps a `declared_name` field (via the extractor's own declaration regex on the same comment-masked text) into each layer function entry; the batch-prompt caller-expectation lookup and the undocumented-submodules advisory both consume this stamped field instead of independently re-deriving "is this a dedup alias" — closing a gap where scoped Chisel declarations (`private[chisel] object Foo`) were missed by a weaker ad-hoc regex.

### Test-file filters and extensions
`.sc`/`.svh` (already accepted for extraction) are now also recognized by the Chisel/Verilog test-file name filters and the two setup-prompt docs, and `build_ext_to_lang` unconditionally covers the whole Chisel/Verilog extension universe once the language token is present.

## Test plan
- [x] 9 standalone test files (`uv run python tests/test_*.py`), all passing — not included in this PR per repo convention on `tests/`
- [x] Real end-to-end runs on riscv-mini (Chisel + generated Verilog) and hand-written SERV RTL, with both `gpt-5.5` and `deepseek-chat` as the OpenCode backend model — 6/6 runs completed successfully (49/49, 14/14, 18/18 modules specced), including a DeepSeek+SERV combination that previously failed to converge before the retry-feedback mechanism landed
- [x] Repeated the DeepSeek run against the final state of this PR's commits — 3/3 designs still complete (45/45, 14/14, 18/18), no regressions from the accumulated hardening

### DIFF
--- a/main.py
+++ b/main.py
@@ -125,7 +125,7 @@ def _has_source_code(proj_dir):
     return False
 
 
-_HARDWARE_LANGUAGES = {"chisel", "verilog", "systemverilog"}
+_HARDWARE_LANGUAGES = {"chisel", "verilog", "systemverilog", "system_verilog"}
 
 
 def _reject_hardware_languages(phases_path):
@@ -504,7 +504,7 @@ if __name__ == "__main__":
     hdl.add_argument(
         "--verilog",
         action="store_true",
-        help="With --hardware: treat the design as Verilog/SystemVerilog (.v/.sv) "
+        help="With --hardware: treat the design as Verilog/SystemVerilog (.v/.sv/.svh) "
         "and generate specs via verilog_spec_generator.",
     )
     parser.add_argument(
--- a/main.py
+++ b/main.py
@@ -518,6 +518,11 @@ if __name__ == "__main__":
     if (args.chisel or args.verilog) and not args.hardware:
         parser.error("--chisel/--verilog select the HDL for --hardware runs; "
                      "add the --hardware flag")
+    if args.resume and not args.hardware:
+        # run_pipeline ignores --resume and starts by WIPING fm_agent/ — the
+        # exact workspace --resume promises to preserve.
+        parser.error("--resume only applies to --hardware runs; "
+                     "add --hardware (and --verilog for Verilog designs)")
 
     start_time = time.time()
     if args.hardware and args.verilog:
--- a/main.py
+++ b/main.py
@@ -7,7 +7,7 @@ from config import (
 )
 from src.file_utils import collect_file_names, is_file_ready
 from src.verification import streaming_reasoner
-from src.extract import run_extraction, EXT_TO_LANG
+from src.extract import run_extraction, EXT_TO_LANG, load_units_manifest
 from src.generate_topdown_layers import generate_topdown_layers
 from src.opencode_trace import (
     finish_opencode_trace,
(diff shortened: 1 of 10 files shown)


haoran-ding (approved):
Thank you very much for the contribution.

PR #96, opened 2026-07-10
Fix json llm output

### LINKED ISSUE #95: Feat: Change the output form of LLM to be json and parse it with json lib
Now FM-Agent ask LLM to return desired information wrapped within [MARKER] ... [MARKER]. Then, it will check the output and parse it based on this form. However, json would be better than such form. Change the prompt to ask LLM to return json, validate whether the output is legal json with required fields, and use lib to parse the json into desired information.

gesi519:
Fixes #95

## Summary
- Replace marker-wrapped direct LLM outputs with strict JSON responses.
- Validate required fields and types for post-condition, incremental-selection, and spec-update responses.
- Reject prose/Markdown-wrapped JSON and retry with the required schema.


haoran-ding (commented):
Why do you push a markdown about erlang here?

gesi519:
I accidentally created this PR from a branch that already contained that earlier commit. I will update the PR from the current branch with only the JSON-output changes. Sorry for the noise.

haoran-ding (changes requested):
There was existing function called _load_spec_check_json in prompts.py, which is more flexible and can resolve cases where LLM outputs some tests with a json. I think it should not be directly changed to use json.loads(text). The function should have the ability to parse the LLM output where there is some other text outside json but there is only one json and the json is legal. This may reduce the number of re-invoking LLM.

haoran-ding (changes requested):
The code may need to be further polished. For example, I think _parse_json_response has the same semantics as _load_spec_check_json. Please check whether there are similar problems and fix them. Thanks :-)

gesi519:
I refactored the duplicated LLM CLI command construction into `build_llm_cli_command()` in `src/llm_client.py`.

haoran-ding (approved):
Approved

PR #98, opened 2026-07-11
sanitize all path-unsafe chars in canonicalize

### LINKED ISSUE #66: C++ operator-overload function names (operator/) crash extraction

tanbing117:
Expands canonicalize() from a /-only workaround to a proper cross-platform filename sanitizer that replaces every character illegal on Windows or Unix (\ / : * ? " < > |) with _.

Fixes \#66 for Windows environments where operator<<, operator|, operator* etc. would still crash on write.

### DIFF
--- a/src/languages/codegraph.py
+++ b/src/languages/codegraph.py
@@ -11,33 +11,34 @@ REGISTRY in src/languages/registry.py. No other files need to change.
 import hashlib
 import logging
 import os
+import re
 import shutil
 import sqlite3
 import subprocess
 from collections import defaultdict
 
 
-_SAFE_REPLACE = str.maketrans({"/": "_"})
-_UNSAFE = set("/")
+_UNSAFE_PATH_RE = re.compile(r'[\\/*?:"<>|]')
 
 
 def canonicalize(func_name):
     """Return a filesystem-safe, FQN-safe version of a function name.
 
-    C++ operator overloads like ``operator/`` contain ``/`` which breaks both
-    file paths and ``::``-separated FQNs.  This function sanitises those
-    characters so the name is safe everywhere it appears: extracted-function
-    file names, FQNs, call-edge keys, and scope.py rankings.
+    C++ operator overloads like ``operator/``, ``operator<<``, ``operator|``
+    contain characters that are illegal in filenames on one or more platforms
+    (``/`` on Unix, ``< > : " | ? *`` on Windows).  This function replaces every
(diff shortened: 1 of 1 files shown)


haoran-ding (commented):
It is unnecessary to consider symbols on windows now. Because FM-Agent only supports linux now.

PR #100, opened 2026-07-12
feat(fqn): qualify C++ member functions with their class name

### LINKED ISSUE #97: Function extraction should specify class name for member functions in class
See this program.

[main.cpp](https://github.com/user-attachments/files/29924665/main.cpp)

There are 4 `action` functions in this file.

In gdb, compiler, etc., class name will be used as function identification: `action`, `Class1::action`, `Class2::action`, `Class3::action`.

We should follow the rule too, otherwise the following execution will be misleading:

```
Success:
uv run python main.py ../test/ --entry-func main-cpp::action
uv run python main.py ../test/ --entry-func main-cpp::Class2::action
uv run python main.py ../test/ --entry-func main-cpp::main --end-func main-cpp::Class2::action
Fail:
uv run python main.py ../test/ --entry-func main-cpp::main --end-func main-cpp::Class1::action
main do not call Class1::action, cannot build call chain.
```

Dshuishui:
## Motivation — resolves #97

Member functions used to get an opaque line-order suffix in their FQN
(`main-cpp::action_1`), so entry-func could not name a specific class's method
and same-name methods across classes were indistinguishable to the user. gdb,
compilers, and humans all identify member functions as `Class::method`.

This makes FM-Agent do the same: `Class2::action`, `LocalStorage::Flush`, etc.
Free functions keep their bare name; the numeric suffix (`_1`, `_2`) is now
kept **only** for true overloads (same class + name, different parameters —
which codegraph's `qualified_name`/`signature` cannot yet separate).

## What changed

- **codegraph**: derive the FQN / extracted-file identifier / spans from
  codegraph's `qualified_name`, canonicalizing each component so a class-scoped
  operator (`Store::operator/`) stays path/FQN-safe. Falls back to the bare name
  for free functions and any unexpected `qualified_name` shape.
- **extract**: write a class-qualified identifier as a flat file that keeps the
  `::` in its name (`Class2::action.cpp`), so `_file_to_fqn` rebuilds
  `...-cpp::Class2::action`.
- **generate_topdown_layers**: collect extracted functions with `os.walk`.
- **entry pipeline**: reverse-map the extracted file to its source, and trim
  by the class-qualified identifier so same-name methods aren't collapsed.

The FQN is produced by the *same* rule in the call edges and the on-disk
extracted files, so the call graph, trimming and entry-func addressing stay
consistent.

## Note — member-function FQN change

This is a sizeable change to how member functions are named. Member-function
FQNs change from bare (`…local_storage-cpp::Flush`) to class-qualified
(`…local_storage-cpp::LocalStorage::Flush`). Any hard-coded FQN downstream must
add the class name (e.g. saved `--entry-func` / `--end-func` arguments). Free
functions are unaffected.

## Verification

Using the example from #97 — one file with four `action` functions (a global
`action` plus `Class1::action`, `Class2::action`, `Class3::action`), where
`main` calls only `Class2::action`:

- Extraction now yields distinct, class-qualified functions:
  `main-cpp::action` (global), `main-cpp::Class1::action`,
  `main-cpp::Class2::action`, `main-cpp::Class3::action` — each specced and
  reasoned about independently.
- `main.py <proj> --entry-func main-cpp::main --end-func main-cpp::Class2::action`
  builds the chain and keeps exactly `main` + `Class2::action.cpp`.
- `main.py <proj> --entry-func main-cpp::main --end-func main-cpp::Class1::action`
  correctly fails — `main` does not call `Class1::action`, so no chain exists.
  Before this change the class name could not be given at all, so the two were
  indistinguishable.

Overloads (`Store::Flush` twice) still fall back to `Store::Flush` /
`Store::Flush_1`; `Store::operator/` becomes the safe `Store::operator_`.

### DIFF
--- a/src/entry_reasoning_pipeline.py
+++ b/src/entry_reasoning_pipeline.py
@@ -87,11 +87,11 @@ def _extracted_file_to_source_rel(extracted_rel):
     Inverse of the extraction layout: ``src/engine/loader-cpp/loadData.cpp``
     (a function file) -> ``src/engine/loader.cpp`` (the source file). Extraction
     builds the function directory by replacing the source filename's last dot
-    with a hyphen (``loader.cpp`` -> ``loader-cpp``). Member functions add a class
-    subdirectory below that (``.../loader-cpp/MyClass/method.cpp``), so we locate
-    the ``<base>-<ext>`` directory itself — scanning from the right so a class
-    dir (which never ends in ``-<known extension>``) is skipped — rather than
-    assuming it is the function file's immediate parent.
+    with a hyphen (``loader.cpp`` -> ``loader-cpp``). Member functions keep the
+    class qualifier in the flat filename (``.../loader-cpp/MyClass::method.cpp``),
+    so the ``<base>-<ext>`` directory is the function file's immediate parent; we
+    still locate it by scanning the path components from the right (matching a
+    component that ends in ``-<known extension>``) so the mapping is robust.
     """
     parts = extracted_rel.split(os.sep)
     for i in range(len(parts) - 2, -1, -1):          # skip the trailing func file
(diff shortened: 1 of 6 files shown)


Dshuishui:
The three findings on the earlier commits (`9b065ca`/`f274b21`) — nested extract paths, stem-only index, and qualifier-only reconciliation — are already addressed on `b26a264` by the flat `Class::method.cpp` layout, a dual-key (qualified + bare) index, and reconciling all `phases.json` sources.

The remaining `_resolve_callee_fqns` point is valid but is a safe over-approximation (a same-named callee may be re-verified once extra — never skipped, so precision only); now tracked in #122 and deferred to a follow-up PR to keep this one focused.

haoran-ding (approved):
Thanks

Issue #104, opened 2026-07-12
Support chisel with circt pass


haoran-ding:
Add support for chisel language by writing a pass for circt to identify modules and construct call graph. Need a new file called chisel.py under src/languages and register functions into registry.py.

PR #106, opened 2026-07-13
Add opt-in --chisel-modules-only filter for Chisel spec generation

Buyi-wsgzg:
## Summary

Adds an opt-in, default-off CLI flag `--chisel-modules-only` that restricts Chisel hardware-spec generation to classes that transitively extend a Chisel hardware base class (`Module`, `RawModule`, `ExtModule`, `BlackBox`, `MultiIOModule`), plus a standalone extractor fix for curried parameter lists that benefits all Chisel runs regardless of the flag.

Two commits, independently reviewable:

1. **`Extract full Chisel declarations with curried parameter lists`** — `_unit_end` truncated the rocket-chip-ecosystem-standard shape `class Foo(a: A, ...)\n  (implicit p: Parameters) extends Module { ... }` after the first parameter list, silently feeding syntactically-broken source to spec generation (found on real Gemmini source: ~30 units affected). The signature walk now follows Scala 2's `ClassParamClause ::= [nl] '(' ...` grammar, operates on comment-masked text (nested comments, `//` inside block comments, `*/`-closing lines), and masks each file once (a per-declaration re-mask was O(n²): 2000 declarations took ~16s, now 0.06s). This fix affects the default Chisel flow and is cherry-pickable on its own.

2. **`Add opt-in --chisel-modules-only filter for Chisel spec generation`** — the feature itself.

## Motivation

Chisel extraction treats every top-level Scala declaration as spec-worthy. On riscv-mini, 32 of 49 extracted units are not hardware (IO Bundles, constant objects, Config classes, the `Main` entry point) — each gets a full LLM spec pass, making Chisel cost ~4x Verilog for the same design, and each shows up as a false positive in the undocumented-submodules advisory (23 entries, most of them noise nobody reads).

## Design

Classification is a text-only heuristic — no Scala import/package resolution — built on one hard rule: **silently dropping a real module is strictly worse than keeping noise.** Every unresolvable case classifies conservative-True (kept):

- unknown external bases (`unresolved_parent`)
- ambiguous same-name candidates (`ambiguous_conservative_true`)
- cyclic or stack-exhausting inheritance chains
- qualified references whose local resolution can't be verified (only a local **True** is trusted; a local False falls back to kept)
- bare root names can't be shadowed into False by unrelated local classes

Mechanics:

- `chisel_decl_info` in `chisel_support.py` is the single source of truthfor `(kind, name, parent, parent_prefix)`; the duplicate declaration regex in `generate_topdown_layers.py` is removed.
- An FQN-indexed, kind-aware, cycle-safe inheritance closure stamps `is_module`/`module_classification_reason` on every Chisel unit unconditionally (flag off = metadata only + a one-time "would exclude" listing).
- Batch generation filters `is_module=False` units; `all_funcs`/ `func_to_layer` stay unfiltered so caller lookups degrade gracefully. Field presence and types are validated loudly, never truthiness-guessed.
- A zero-post-filter-modules guard fails loudly (scoped to the topdown files actually written this run) instead of quietly emitting nothing.
- The flag survives `--resume` (subprocess argv + the recursive setup-rerun self-call).
- The shared undocumented-submodules advisory gains an explicit `filter_non_modules` parameter — Chisel call site only; the Verilog call site is byte-for-byte unaffected.
- Stale specs left by a prior unfiltered run are warned about, never deleted (matches the repo's existing no-cleanup convention).

## Validation

Real DeepSeek runs (same source, same backend, on/off):

| Run | Completion | Spec calls | Advisory |
|---|---|---|---|
| riscv-mini off | 49/49 units | 78 | 23 |
| riscv-mini **on** | **17/17 true modules** | **31 (−60%)** | **7 (−70%)** |
| gemmini systolic-array subset off | 10/10 units | 18 | 3 |
| gemmini subset **on** | **4/4 true modules** | **5 (−72%)** | **0** |

All 16 advisory entries eliminated on riscv-mini were known false positives; one surviving entry exposed a real caller-contract gap (a malformed `# Submodule:` heading breaking the caller-expectation handoff) that had been buried in prior runs' noise.

Non-regression: a real Verilog DeepSeek run (14/14 modules, call count on par with the pre-change baseline, zero advisory) and a deterministic software-pipeline smoke (no `is_module` leakage, embedded `[SPEC]` format intact). Developed test-first; the accompanying test suite can be contributed alongside if useful.

## Relationship to #104 (Support chisel with circt pass)

Complementary, not competing. Elaboration via CIRCT gives a compiler-grade module set and instance graph, but requires a buildable project (JDK/sbt/deps/a Top+Config) and only sees the design one configuration selects — riscv-mini's own CIRCT-generated Verilog contains 14 modules: the config-unselected implementations (`AluSimple`, `BrCondSimple`, `ImmGenMux`) don't exist in it, while 3 generated memory modules in it have no Scala source to attach a spec to. This text-only layer covers what elaboration structurally can't: repos that don't build, partial source trees, and all configuration variants, with zero toolchain requirements.

If #104 lands, its natural role in this flow is an **oracle**: cross-check `is_module` and provide precise instance edges (potentially hardening the Chisel advisory into a gate like Verilog's) when a build is available, reporting source declarations not covered by elaboration rather than silently skipping them. The classification metadata this PR stamps on every unit is the interface such an oracle would plug into.

## Known limitations (documented in README, pinned by regression tests)

- Bare `Bundle`/`Record`/`Data` are interpreted as chisel3's (imports are discarded at extraction); an import alias or an external class sharing one of the fixed bare names can be deterministically misclassified.
- `object`/`def` declarations are never eligible for module-ness (Scala singleton semantics: not instantiable via `Module(new ...)`, breaks re-elaboration).
- Scala 2 (2.12/2.13) syntax only; Scala 3 significant-indentation and the deprecated early-initializer syntax are unsupported input forms.

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,6 +137,7 @@ uv run python main.py <proj_dir>
 | `--chisel`   | With `--hardware`: treat the design as Chisel (Scala). This is the default HDL, so `--hardware` alone is equivalent to `--hardware --chisel` |
 | `--verilog`  | With `--hardware`: treat the design as Verilog/SystemVerilog (`.v`/`.sv`/`.svh`) |
 | `--resume`   | Resume an interrupted `--hardware` run: reuse `groups.json` and only regenerate missing module specs |
+| `--chisel-modules-only` | With `--hardware --chisel`: skip spec generation for Chisel classes confidently identified as non-hardware (IO Bundles, constant objects, and similar), keeping units that transitively extend `Module`/`RawModule`/`ExtModule`/`BlackBox`/`MultiIOModule`. Classes whose module-ness can't be determined heuristically (unresolved external bases, ambiguous or cyclic inheritance) are conservatively kept, not excluded. This is a text-only heuristic with no real Scala import/package resolution, so an unrelated class elsewhere in the project sharing a parent's bare name can, in rare cases, still cause a misclassification. An import alias that renames a base to `Module`/`RawModule`/`ExtModule`/`BlackBox`/`MultiIOModule`/`Bundle`/`Record`/`Data` (e.g. `import chisel3.{Module => Bundle}`) is treated as if it named that class directly, which can deterministically misclassify a real module. Extraction is unaffected; this only filters spec generation. |
 
 ### Generating Specs for Hardware Designs (`--hardware`)
 
@@ -144,6 +145,12 @@ For Chisel (Scala) hardware designs:
 
 ```bash
 uv run python main.py <proj_dir> --hardware
+```
+
+To skip spec generation for non-hardware Chisel units (IO Bundles, constant objects, `Main` entry points):
+
+```bash
+uv run python main.py <proj_dir> --hardware --chisel-modules-only
 ```
 
 For Verilog/SystemVerilog hardware designs:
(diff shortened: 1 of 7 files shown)


haoran-ding (approved):
Approved. Thank you very much for the contribution.

PR #116, opened 2026-07-14
Feat/rebuild structured function metadata

### LINKED ISSUE #105: Store [SPEC] block and [INFO] block in another two separate json files
Now the spec and info blocks of each extracted function are in the same file of the function implementaion. Now store them into another two seaparate json files near the function file.

QiaoyiZheng:
## Summary

Move each extracted function's behavioral specification and callee metadata out of the function source file and into two adjacent JSON sidecar files:

- `<function-file>.spec.json`
- `<function-file>.info.json`

The extracted function file now contains only the original function implementation and is no longer overwritten during specification generation.

Closes #105.

## Changes

- Define structured JSON formats for function specifications and callee expectations.
- Generate `.spec.json` and `.info.json` next to every extracted function.
- Require both sidecars to exist and strictly match the supported schemas before a function is considered ready.
- Exclude metadata sidecars from function collection and top-down layer generation.
- Read caller specifications and callee expectations from JSON sidecars.
- Rebuild the existing reasoner-facing spec text and `FunctionSpecMap` at the parser boundary.
- Update incremental reasoning to:
  - preserve sidecars when function sources are re-extracted;
  - generate or update structured metadata objects;
  - write only sidecar files;
  - remove sidecars for deleted functions;
  - reconcile caller `.info.json` files without modifying caller source files.
- Support only the new JSON sidecar format; legacy inline `[SPEC]` and `[INFO]` blocks are not migrated automatically.

## JSON formats

`<function-file>.spec.json`:

```json
{
  "unit": "<file path relative to repo root>",
  "signature": "<FunctionName>(<params>) -> <ReturnType>",
  "pre_condition": "<what must hold before the call>",
  "post_condition": "<what the function guarantees after return>"
}


haoran-ding (changes requested):
"name" and "unit" field of generated spec.json and info.json seem unnecessary. Do you think they can be removed?

haoran-ding (commented):
Why do you add " and not _is_metadata_sidecar(fname):" in generate_top_down_layers.py? The only modification to pipeline_setup.py is an empty line, which is wierd.

QiaoyiZheng:
> "name" and "unit" field of generated spec.json and info.json seem unnecessary. Do you think they can be removed?

I agree that unit in spec.json is redundant because the sidecar’s location already identifies the source unit, so we can remove it. However, name in each info.json callee entry is currently used to associate the expectation with a callee in the call graph and to propagate specification changes. The signature field is LLM-generated, language-dependent display text rather than a canonical identifier. It may change when parameters or return types change, and identical signatures may exist in different classes, namespaces, or files. Therefore, removing name would require another stable identifier, such as a fully qualified function ID. I suggest removing unit while retaining name.

QiaoyiZheng:
> Why do you add " and not _is_metadata_sidecar(fname):" in generate_top_down_layers.py? The only modification to pipeline_setup.py is an empty line, which is wierd.

The extracted-functions directory now contains both extracted source files and their adjacent .spec.json/.info.json sidecars. _collect_phase_files() walks all regular files in that directory, so during resume or layer regeneration it would otherwise treat the sidecars as function source files and include them in layer construction. The metadata-sidecar check ensures that only extracted functions are collected.
You are right about pipeline_setup.py. The blank line was an accidental leftover from separating an unrelated source_files fix from this branch. I have removed it, so pipeline_setup.py is no longer part of the PR diff.

QiaoyiZheng:
The changes have been made and have passed testing. Please review.

haoran-ding (changes requested):
The generated files, such as fm_agent/spec_prompts/*.txt. still have something stale like "schema_version". Please check why and fix it.

QiaoyiZheng:
> The generated files, such as fm_agent/spec_prompts/*.txt. still have something stale like "schema_version". Please check why and fix it.



> Thanks for catching this. The stale `schema_version` was coming from old `batch_*.txt` files that remained in `fm_agent/spec_prompts/` after the generator moved to another layer or produced a different batch set. The manifest was updated, but prompt files no longer belonging to the current generation were not removed.
>
> I fixed the generator so that each actual generation removes obsolete `batch_*.txt` files before writing the current batch prompts.
>
> I also ran the full pipeline and verified that:
>
> - the generated batch files match the current manifest;
> - no stale layer or batch prompt remains;
> - no file under `fm_agent/`, including generated scripts and caches, contains `schema_version`;
> - generated `.spec.json` and `.info.json` files use only the current schema fields.
>
> Fix: `5410bc5 fix(spec): remove stale generated batch prompts`
>
> please review the updated change again

haoran-ding:
Where is your new commit?

QiaoyiZheng:
> Where is your new commit？

sorry i forgot to push. it is able to be reviewed now

haoran-ding (changes requested):
I think you did not find the real root cause. If you search for schema_version, you will find  many files containing it. It is the real cause that should be fixed.

haoran-ding (approved):
Thanks

PR #120, opened 2026-07-14
Pin codegraph to the fmagent fork's fixed release

### LINKED ISSUE #119: install.sh installs the latest upstream codegraph unpinned, which can silently break C function extraction
## Summary

FM-Agent's `install.sh` installs codegraph via `bun install -g @colbymchenry/codegraph` with **no version pin**, so every setup pulls whatever upstream currently ships as `latest`. The current latest (v1.4.1) contains a C extraction bug: a function shaped `MACRO_ATTR TYPEDEF_RET name(VOID)` (e.g. OpenHarmony LiteOS-M's `LITE_OS_SEC_TEXT_INIT UINT32 LOS_KernelInit(VOID)`) is indexed under its parameter list `(VOID)` instead of its name. Such functions silently drop out of the call graph.

## Impact

- C functions of this shape — common in OS/RTOS kernels — are indexed as `(VOID)` and disappear from the extracted set and the call graph.
- In entry-func mode this can fail the reachability check with `ValueError: none of the requested end_funcs are reachable ...` when a requested or intermediate function is one of the lost ones.
- For ordinary runs the failure is **silent** — extraction just misses functions and coverage drops with no error.

## Root cause

- **No version pinning** — FM-Agent tracks upstream codegraph `latest`, so an upstream regression lands automatically on the next `install.sh` run.
- Upstream fix (colbymchenry/codegraph#1211) is still open / unreleased; the latest release still has the bug.

## Proposed fix

- Pin FM-Agent to a maintenance fork's fixed release: **`fmagent-project/codegraph` `v1.3.0-fmagent.1`** (upstream v1.3.0 + the #1211 fix), installed by `install.sh` as a self-contained prebuilt bundle (no Node/build).
- Add a **one-line version switch** so bumping the pin is trivial.
- **Version-check** on install (not mere presence) so a stale/wrong codegraph is corrected.
- For existing users who already have the official codegraph on PATH, print a **prominent, actionable warning with the exact removal command** — never auto-uninstall.

Dshuishui:
Closes #119.

## Summary

Makes FM-Agent reliably use our maintenance fork's fixed codegraph, both at install time and at run time.

**`install.sh` — pin the install:**
- **Pin + switch** — installs `fmagent-project/codegraph` `v1.3.0-fmagent.1` (upstream v1.3.0 + the C macro-attribute name-loss fix). Bump the one `CODEGRAPH_VERSION` line to switch versions.
- **Self-contained bundle** — installed via the fork's `install.sh` (a prebuilt per-OS bundle; no Node or build tools needed).
- **Checks/installs against `$CODEGRAPH_BIN_DIR`** (default `~/.local/bin`) specifically, and verifies the launcher by file — so a bin dir that is not on the caller's PATH does not abort setup, and that dir always holds the pinned version.

**`src/languages/codegraph.py` — resolve it at run time:**
- The pinned build lands in `$CODEGRAPH_BIN_DIR/codegraph`. FM-Agent now invokes it by **absolute path** rather than a bare `codegraph` resolved against PATH — so the pinned build is used even when that dir is not on PATH (the macOS default), and an older codegraph earlier on PATH cannot shadow it (the pipx/Playwright pattern of running a managed binary from its known location). Falls back to a bare `codegraph`, then the regex extractor, when the pinned build is absent.

## Why

Tracking upstream `latest` silently pulls in a C extraction bug: a function shaped `MACRO_ATTR TYPEDEF_RET name(VOID)` is indexed under its parameter list `(VOID)` instead of its name, so such functions drop out of the call graph. Details in #119.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -145,12 +145,46 @@ else
     bunx oh-my-openagent install --no-tui --claude=no --gemini=no --copilot=no
 fi
 
-# ---------- codegraph ----------
-if command -v codegraph &>/dev/null; then
-    echo "[ok] codegraph found: $(codegraph --version 2>/dev/null || echo 'unknown version')"
+# ---------- codegraph (pinned to fmagent fork) ----------
+# Pin to our maintenance fork's fixed build: the latest upstream release still
+# has a C extraction bug (a macro attribute + typedef'd return type on a no-arg
+# function is indexed under its parameter list `(VOID)` instead of its name), so
+# tracking upstream `latest` silently drops functions. The fork ships a
+# self-contained prebuilt bundle (no Node/build). Change CODEGRAPH_VERSION to switch.
+CODEGRAPH_REPO="fmagent-project/codegraph"
+CODEGRAPH_VERSION="v1.3.0-fmagent.1"                 # <- switch: bump this one line
+codegraph_want="${CODEGRAPH_VERSION#v}"
+if codegraph --version 2>/dev/null | grep -qx "$codegraph_want"; then
+    echo "[ok] codegraph $codegraph_want already installed"
 else
-    echo "[..] installing codegraph"
-    bun install -g @colbymchenry/codegraph
+    echo "[..] installing codegraph $CODEGRAPH_VERSION from $CODEGRAPH_REPO"
+    curl -fsSL "https://raw.githubusercontent.com/$CODEGRAPH_REPO/main/install.sh" \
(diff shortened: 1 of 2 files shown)


Dshuishui:
Superseded by #125 (merged), which moves the codegraph pin into `fm-agent.toml`'s `[codegraph]` section (same fork + `v1.3.0-fmagent.1`), reads it in both `install.sh` and `_codegraph_cmd()`, and adds a runtime version-mismatch warning. Closing.

PR #125, opened 2026-07-15
refactor(config): consolidate runtime settings into fm-agent.toml + a typed loader

### LINKED ISSUE #124: Runtime configuration is scattered across multiple files with no single source of truth
## Problem

FM-Agent's configuration is spread across several places with no single source of truth, and secrets are mixed in with ordinary tunables.

Current state:

- **`config.py`** is *intended* as the central config hub (LLM/model/scope knobs, read via `os.environ.get` with defaults), but it isn't the only reader.
- **`.env`** mixes a real secret (`LLM_API_KEY`) with non-secret runtime settings (`LLM_MODEL`, `LLM_API_BASE_URL`, `FM_AGENT_MODEL_BACKEND`, `OPENCODE_MODEL_PROVIDER`, `ELP_*`, …). Users editing a model name touch the same file that holds their API key.
- **`load_dotenv()` is called twice** (in `config.py` and in `src/cli_backend.py`), and `FM_AGENT_MODEL_BACKEND` / `LLM_EFFORT` are re-read directly in `cli_backend.py` instead of coming from `config.py`.
- **Several modules read their own env directly**, bypassing the hub: `src/languages/erlang.py` (`ELP_COMMAND`, `ELP_TIMEOUT_SECONDS`), `src/llm_client.py` (`INJECT_ID`, `INJECT_HOST`), `src/domain_knowledge.py` (`FM_AGENT_DOMAIN_KNOWLEDGE`).
- **Some tunables are hardcoded** in `config.py` with no override path (`MAX_WORKERS`, `MAX_SPC_ITER`, `GRANULARITY`, `OPENCODE_MAX_RETRIES`).
- **The codegraph version is pinned only inside `install.sh`**, so the Python side and the installer have no shared, declarative source for "which codegraph build this project expects."

## Why it matters

- No single place to see or change what the agent runs with.
- Secrets and shareable defaults live in the same untracked file, which is easy to leak and impossible to commit sane defaults for.
- New contributors have to grep multiple files to learn the knobs.
- The install script and the runtime can drift on values they both care about (e.g. the codegraph version).

## Proposed direction

Adopt the standard layered-config pattern used by most modern Python apps:

```
code defaults  <  fm-agent.toml (committed, non-secret)  <  .env / env vars
```

1. Add a committed **`fm-agent.toml`** holding all non-secret defaults (model, backend, provider, scope knobs, runtime limits, erlang, and a placeholder for the pinned codegraph version).
2. Keep **`.env` for secrets only** (`LLM_API_KEY`).
3. Turn **`config.py` into a single typed loader** (via `pydantic-settings`) that merges the toml file, `.env`, and process env with a clear precedence, and validates types. Existing `from config import X` imports keep working.
4. **Route the scattered per-module env reads** for *configuration* through the loader (erlang, injection, domain-knowledge). Environment-*detection* reads (host/CLI markers such as `CODEX_HOME`, `CLAUDE_PLUGIN_ROOT`) stay as-is — they inspect the ambient runtime, not user config.

Env-var overrides are preserved, so existing `.env` setups keep working with no migration step.

### LINKED ISSUE #126: install.sh Python floor (3.10) inconsistent with pyproject requires-python (>=3.12)
`install.sh` gates on Python `3.10+`, but `pyproject.toml` declares `requires-python = ">=3.12"`. A 3.10/3.11 interpreter passes the installer's check, then `uv sync` rejects it as incompatible with `>=3.12` — so the version gate lets through a Python the project can't actually run on.

Fix: align the `install.sh` check (and the README requirement) to `>=3.12`.

Addressed in #125.

Dshuishui:
## What

Implements the direction in #124: one committed, non-secret config file plus a single typed loader, with secrets kept in `.env`.

- **Add `fm-agent.toml`** — committed, non-secret defaults for `[llm]`, `[runtime]`, `[scope]`, `[erlang]`, and `[codegraph]` (the pinned codegraph repo / version / bin_dir).
- **Rewrite `config.py`** as a single `pydantic-settings` loader. Precedence: `process env > .env > fm-agent.toml > built-in defaults`. Values are type-validated. Every existing `from config import *` constant is preserved as a module-level alias (now pinned by an explicit `__all__`).
- **Slim `.env.example`** to secrets only (`LLM_API_KEY`), noting any setting can still be overridden via its env var.
- **Route the scattered per-module config reads** through the loader: `erlang.py` (`ELP_*`), `llm_client.py` (`INJECT_*`), `domain_knowledge.py` (`FM_AGENT_DOMAIN_KNOWLEDGE`), and `cli_backend.py` (`FM_AGENT_MODEL_BACKEND`, `LLM_EFFORT` — also drops its redundant second `load_dotenv()`). Environment-*detection* reads (host/CLI markers) intentionally stay as-is.
- **Make `fm-agent.toml` the single source for the OpenCode provider.** Previously the user had to hand-write (and keep in sync) a `provider` block in `~/.config/opencode/opencode.json`. FM-Agent now builds that block from `[llm]` (provider / base_url / name / new `api_style`) and injects it into the OpenCode subprocess via `OPENCODE_CONFIG_CONTENT` — OpenCode's highest-precedence config source, so it overrides a same-named provider in the user's `opencode.json` while deep-merge preserves the rest (other providers, the `plugin` array). The key stays as `{env:LLM_API_KEY}`, never written to disk; injection only happens when the `[llm]` fields and `LLM_API_KEY` are all set.
- **Drive codegraph from `fm-agent.toml` `[codegraph]` too.** `install.sh` reads `repo` / `version` / `bin_dir` from the toml and installs the pinned fork build there (verifying the installed version, not just presence); the runtime invokes it by absolute path from that same `bin_dir`, so PATH order/absence can't select a different build, and warns (without failing) when the running version doesn't match the pin. Bumping `[codegraph].version` switches the codegraph build in one place.
- **Add `pydantic-settings>=2.2`** (+ `uv.lock`).
- **Align the Python floor to `pyproject`'s `requires-python` (`>=3.12`)** in `install.sh` and `README` — the new `tomllib` import needs 3.11+, and `install.sh`/`README` previously claimed 3.10 while `pyproject` already required 3.12 (#126).

## Compatibility

No migration needed — every legacy env var still overrides, so existing `.env` files keep working. The env->setting map lives in one table in `config.py`.

Two intentional behavior changes worth calling out:
- **Fail-fast on bad config.** An invalid value (e.g. a non-integer `ELP_TIMEOUT_SECONDS`) now stops startup with a readable message, instead of the old warn-and-fall-back that only triggered when Erlang analysis ran.
- **Typos are rejected.** Unknown keys or sections in `fm-agent.toml` raise (`extra="forbid"`) rather than being silently ignored.
- `MAX_SPC_ITER`, `GRANULARITY`, `MAX_WORKERS`, `OPENCODE_MAX_RETRIES` were hardcoded before; they are now overridable via the toml/env. Defaults are unchanged.

## Out of scope (follow-ups)

- An interactive `fm-agent init` command.

## Verification

- `config` and all 13 config-consuming modules import clean; all 21 back-compat constants present with values identical to the old `config.py`.
- Precedence confirmed (real env > .env > toml > default); legacy env vars override the toml with type coercion; an invalid value or a typo'd toml key/section fails fast at startup.
- OpenCode injection verified against a live OpenCode install: `opencode models` lists the injected provider/model with no `opencode.json` provider block; `opencode debug config` confirms a same-named provider is overridden by the injected `baseURL` while unrelated providers and the `plugin` array survive the merge.

Closes #124.
Closes #126.


fm-agent.toml
```diff
@@ -0,0 +1,36 @@
...
+[llm]
+name     = "anthropic/claude-sonnet-4.6"     # LLM_MODEL
+provider = "openrouter"                        # OPENCODE_MODEL_PROVIDER — must match a provider in ~/.config/opencode/opencode.json
+base_url = "https://openrouter.ai/api/v1"      # LLM_API_BASE_URL
+backend  = "opencode"                          # FM_AGENT_MODEL_BACKEND — opencode | auto | codex-cli | claude-cli
+effort   = ""                                  # LLM_EFFORT — reasoning-effort hint ("", low, medium, high)
+
+[runtime]
+max_workers                = 10    # MAX_WORKERS
+max_spec_iter              = 5     # MAX_SPC_ITER
```
heiyan-2020, line 18: add comment to describe the usage of this config variable

config.py
```diff
@@ -1,46 +1,253 @@
...
+    version: str = ""
+
+
+class _LayeredSource(PydanticBaseSettingsSource):
+    """``fm-agent.toml`` as the base layer, mapped process env overlaid on top."""
+
+    def __init__(self, settings_cls, path: Path):
+        super().__init__(settings_cls)
+        data: dict = {}
+        if path.is_file():
```
heiyan-2020, line 170: if path resolve failed, should thorw exception, at least print some warning messages rather than silently read default values

install.sh
heiyan-2020, line 1: After moving all config variables to fm-agent.toml, install.sh may not read those values

fm-agent.toml
Dshuishui: Thanks! To avoid duplicating the full descriptions, I moved the parameter table out of the README into _docs/configuration.md_, and added a pointer to it at the top of _fm-agent.toml_.

config.py
Dshuishui: Thanks for the catch! Fixed in 830cd63. Two cases now:
- An explicit `FM_AGENT_CONFIG` pointing at a missing file already fails fast (raises `SystemExit`).
- A missing *default* `fm-agent.toml` now prints a warning to stderr before falling back to the built-in defaults (which are kept identical to the toml), instead of doing it silently.

I also widened the handling so a *syntax error* in the toml (`TOMLDecodeError`) shows the same readable message instead of a raw traceback.

install.sh
Dshuishui: Thanks! `install.sh` only needs two things from the config: `[llm].backend` (to decide opencode vs. a local CLI) and the `[codegraph]` repo/version/bin_dir (to install the pinned binary). The rest are runtime-only and aren't needed at install time.

How: right after `uv sync`, `install.sh` runs a short `python` snippet that does `from config import settings` and prints those values. It no longer parses the TOML itself — the env>toml precedence, `~` expansion, and validation all come from `config.py`, the same loader the runtime uses. Done in 830cd63.

haoran-ding (approved):
Thanks

PR #130, opened 2026-07-15
Add conservative Chisel module-only extraction and CIRCT-backed module graph support

okcomputer2000:
## Summary

This PR adds a conservative Chisel hardware extraction path on top of the current `main` branch.

The first version stays intentionally narrow:

- analysis units: `Module`, `RawModule`, `BlackBox`, `ExtModule`, `MultiIOModule`
- `Bundle` is treated as context only, not as a standalone unit
- graph relations are limited to the module instantiation graph
- source filtering is conservative, and CIRCT-emitted Verilog is treated as authoritative when available

## Approach

The implementation adds a dedicated Chisel path instead of trying to stretch the generic Scala/function extraction flow:

- detect Chisel source files conservatively
- extract only hardware-module units from Chisel/Scala source
- keep Chisel test-file handling path-aware so `src/main/...` production files are not dropped by filename heuristics
- build module-to-module edges from instantiation structure
- when CIRCT-emitted Verilog is available, use it to authoritatively filter the module set and override source-derived edges

This keeps the Chisel path isolated and minimizes coupling with the existing non-Chisel pipeline.

## What is included

- Chisel-specific extraction support
- conservative module classification
- module-instantiation edge construction
- CIRCT sidecar graph loading and parsing
- integration with the existing extraction and topdown-layer pipeline

## What is intentionally not included

- `Bundle` as an independent analysis unit
- broader Scala semantic extraction
- non-module graph relations
- deeper behavior inference beyond conservative module detection and instantiation links

## Validation

Tested locally against public Chisel codebases and local sample trees, including:

- `schoeberl/chisel-examples`
- `ucb-bar/chisel2-deprecated`
- `ucb-bar/chipyard`
  - `fpga/src/main/scala`
  - `generators/chipyard/src/main/scala`
  - `generators/chipyard/src/main/resources/vsrc`

Also validated the relevant FM-Agent pipeline stages locally by running the real
`run_extraction` and `generate_topdown_layers` flow on a local test project, and stopping before any LLM-driven stages.

## Current limitations

- module detection is still intentionally conservative and may skip some valid hardware-like Scala declarations
- CIRCT integration currently depends on emitted Verilog sidecar input rather than a deeper direct FIRRTL/CIRCT integration
- `Bundle` is only preserved as surrounding context inside extracted modules
- graph precision is limited to module instantiation relationships in this version

## Next steps

- expand coverage only where real Chisel code demonstrates missed module forms
- evaluate whether a second version should add more Chisel-specific structure while keeping module-only output as the default
- improve CIRCT-driven validation and fixtures as more real-world examples are collected


haoran-ding (changes requested):
The PR does not use https://github.com/llvm/circt. All code about supporting a new language should be put under src/languages and register the interface in registry.py.

okcomputer2000:
This PR adds Chisel support as a language backend under `src/languages` and registers it in `src/languages/registry.py`.

Current status:
- Chisel module extraction is implemented in `src/languages/chisel.py`.
- A CIRCT pass/plugin is added to identify hardware modules and build a module instantiation graph.
- When CIRCT input is available, CIRCT is treated as the authority for module discovery and module-instantiation edges.
- Python only slices source bodies for the matched Chisel module declarations.
- Supported extracted units are currently `Module`, `RawModule`, `ExtModule`, `BlackBox`, and `MultiIOModule`.
- `Bundle` is kept as context and is not extracted as a standalone unit.

Known limitations:
- For Chisel, the graph here is a module instantiation graph, not a traditional software function call graph.
- The current Chisel/FIRRTL/CIRCT flow reliably provides source start locations, but not stable full source ranges.
- Because of that, the Python side still needs a small amount of source slicing logic to cut module bodies from Scala files.

Validation:
- Tested on local checkouts of `schoeberl/chisel-examples`, `ucb-bar/chisel2-deprecated`, and `ucb-bar/chipyard`.
- Verified the CIRCT-backed extraction path, module graph generation, and `run_extraction` integration.

Next steps:
- Keep CIRCT authoritative for module identification and module-instantiation edges.
- Reduce the remaining Chisel-specific source heuristics further.
- If upstream CIRCT/Chisel can provide stable full source ranges later, simplify the Scala source slicing layer further.

Issue #121, opened 2026-07-15
`_bare_function_name` over-strips overloaded functions, producing wrong identifiers


Dshuishui:
`_bare_function_name` (in `src/languages/codegraph.py`, applied by `_extraction_ident`) strips too aggressively for some overloaded functions: it drops the part of the name after the leading identifier, so distinct overloads collapse to the same wrong identifier and become indistinguishable. C++ operator overloads are the clearest case — every `operator<sym>` in a class is reduced to the bare word `operator`.

### Reproduction

A single class with four distinct operator overloads:

```cpp
struct Vec {
    int x;
    bool operator==(const Vec &o) const;
    int  operator[](int i) const;
    Vec  operator/(int d) const;
    Vec  operator+(const Vec &o) const;
    void normalize();
};
```

Running the pipeline on it extracts (exit 0):

```
Vec::operator.cpp     (operator==)
Vec::operator_1.cpp   (operator[])
Vec::operator_2.cpp   (operator/)
Vec::operator_3.cpp   (operator+)
Vec::normalize.cpp    (control — unaffected)
```

Four distinct operators become `Vec::operator` + opaque line-order suffixes (`_1`/`_2`/`_3`), reintroducing for operators the "can't tell which is which / unstable" problem #97 removed for same-named methods.

### Impact

Every overloaded operator in a class ends up with the wrong, non-distinct identifier, which breaks things that depend on the identifier being correct and stable:

- **entry-func can't address a specific operator.** `--entry-func ...::Vec::operator_2` is meaningless (which operator is `_2`?) and unstable — the suffix follows source line order, so reordering the operators reassigns it.

Not a crash and no functions are dropped (each still gets its own file), but the identifiers are wrong, so any per-function reasoning about an operator can be attributed to the wrong one. Affects classes with multiple operator overloads (common in C++).

Dshuishui:
Resolved by #134 (merged). `_bare_function_name` now preserves C++ operator overload identifiers — symbolic operators (`operator==`, `operator[]`, `operator+`, `operator,`, …), `operator()`, and `operator new`/`delete`(`[]`) — so distinct overloads in a class no longer collapse to a bare `operator` with source-order suffixes.

One rare edge remains (noted on the PR): a function named *exactly* `operatornew`/`operatordelete` (no suffix) still maps to `operator new`/`operator delete`. Unusual naming, left for now. Closing.

PR #134, opened 2026-07-17
fix(codegraph): preserve distinct C++ operator overload identifiers

okcomputer2000:
## Summary

This fixes issue #121 by making `_bare_function_name()` preserve the meaningful part of C++ operator overload names instead of collapsing every `operator<sym>` to the bare word `operator`.

The change replaces the previous broad regex behavior with explicit parsing for operator-prefixed names:
- `operator[]`
- `operator()`
- symbolic operators like `operator==`, `operator/`, `operator+`
- `operator new`, `operator new[]`
- `operator delete`, `operator delete[]`

Non-operator name handling is left unchanged.

## Problem

When codegraph returns C++ method names like:
- `operator==`
- `operator[]`
- `operator/`
- `operator+`

the old `_bare_function_name()` stripped them too aggressively and reduced all of them to `operator`. That forced downstream deduping to invent unstable suffixes like `_1`, `_2`, `_3`, making extracted identifiers ambiguous and source-order dependent.

## Result

After this change, overloaded operators remain distinct through extraction and function-span naming, so downstream identifiers stay meaningful and stable.

Example output now stays differentiated as:
- `Vec::operator==`
- `Vec::operator[]`
- `Vec::operator_`
- `Vec::operator+`

## Validation

- Added inline assertion coverage for operator overload forms and existing non-operator cases.
- Reproduced the original C++ example against a fresh `codegraph` index and confirmed distinct extracted names.

## Note on conversion operators

This change intentionally fixes the operator-overload cases reported in issue #121 for symbolic operators such as `operator==`, `operator[]`, `operator/`, `operator+`, along with `operator()` and `operator new/delete`.

It does **not** fully address C++ conversion operators yet, such as:

- `operator bool()`
- `operator int()`
- `operator const char *()`

In local validation, these still collapse to a shared `operator`-based identifier and are then disambiguated only by the existing `_1`, `_2`, ... suffixing. There is also an upstream limitation from `codegraph` itself: for example, `operator const char *()` may already be stored only as `operator const` in the indexed node name, which means some type information is lost before FM-Agent processes it.

So this PR fixes the concrete issue 121 regression for symbolic operator overloads, but conversion operators remain a separate follow-up area.

### DIFF
--- a/src/languages/codegraph.py
+++ b/src/languages/codegraph.py
@@ -100,6 +100,7 @@ def _bare_function_name(name: str) -> str:
     - Simple identifier: ``"my_func"`` -> ``"my_func"``
     - Qualified name: ``"ns::Cls::method"`` -> ``"method"``
     - Go pointer receiver: ``"(*T).Method"`` -> ``"Method"``
+    - C++ operator overload: ``"Vec::operator=="`` -> ``"operator=="``
     - Function-pointer: ``"(*func)(type *param)"`` -> ``"func"``
     - Pointer return: ``"*func_name(...)"`` -> ``"func_name"``
     - this-dot: ``"this.onClick"`` -> ``"onClick"``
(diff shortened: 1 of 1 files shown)


okcomputer2000:
Addressed the follow-up review note by including the comma operator in the symbolic parser as well. operator, is now preserved like the other overloadable symbolic C++ operators, so comma overloads no longer fall back to ambiguous operator / operator_1 identifiers.

src/languages/codegraph.py
```diff
@@ -109,7 +110,35 @@ def _bare_function_name(name: str) -> str:
...
+    elif "." in tail:
+        tail = tail.rsplit(".", 1)[1].lstrip()
+
+    if tail.startswith("operator"):
+        rest = tail[len("operator"):].lstrip()
+        if rest.startswith("[]"):
+            return "operator[]"
+        if rest.startswith("()"):
+            return "operator()"
+        if rest.startswith("new"):
```
Dshuishui: `startswith("new")`/`startswith("delete")` match by prefix, so distinct methods that merely start with `operatornew`/`operatordelete` collapse to the same id — e.g. a class with `operatornewEntry()` and `operatornewBucket()` both become `operator new`, so their extracted files collide and one is silently dropped (never verified, missing from the call graph). Rare naming, but the failure is silent — worth guarding here.

okcomputer2000:
Fixed by tightening `operator new` / `operator delete` detection to exact matches only, so `operatornew*` and `operatordelete*` methods no longer collapse into the allocator/deallocator identifiers.

Dshuishui (approved):

Dshuishui:
> Fixed by tightening `operator new` / `operator delete` detection to exact matches only, so `operatornew*` and `operatordelete*` methods no longer collapse into the allocator/deallocator identifiers.

Thanks — the exact-match tightening handles the `operatornew*`/`operatordelete*` suffix cases. One narrow residual: a method named *exactly* `operatornew` or `operatordelete` (no suffix) still resolves to `operator new`/`operator delete` and collides with a real `operator new`/`delete` in the same class — after stripping the `operator` prefix, the leftover is exactly `new`, and the space that separates `operator new` from `operatornew` is already gone. The naming is rare, so this can be left for now.

PR #142, opened 2026-07-21
Refactor plugin execution to use Python function interfaces

### LINKED ISSUE #144: 插件接口设计（Part2.1）
Part 1 has been completed. See: https://github.com/fmagent-project/FM-Agent/issues/131

A plugin can replace or modify the input and output of each stage. In Part 1, this was implemented with command-line commands. Now it needs to be changed to Python functions.

0. Remove the required "replace_cmd" and "output_process" fields in Part 1 from plugin.json.
1. Define the parameters and return values for the replacement and modification functions of each stage.
    - Since the input and output of each stage are files, a stage replacement function should take one or more file paths as parameters and return one or more file paths.
    - Each stage should have two modification functions: one for the input file and one for the output file. Each function should take one file path as its parameter and return nothing.
2. Each plugin should include a Python file that implements these functions as needed.
3. When a plugin is loaded, the system should automatically check:
    - whether its plugin.json matches the implemented functions;
    - whether the function parameters and return values are defined correctly.

In this issue, you only need to make these changes for the first two stages.

tanbing117:
Closes #144

### Motivation

Previously, plugin stage customization relied on shell commands declared in `plugin.json`. While functional, this design required launching subprocesses, offered limited validation, and tightly coupled plugins to command-line execution.

This PR introduces a Python function-based plugin API, enabling plugins to be loaded as modules, validated at startup, and invoked directly from the pipeline.

### What changed

- Remove the `replace_cmd` and `output_process` fields from `plugin.json`
- Introduce `plugin.py` as the implementation entry point for plugins
- Define standard hook interfaces for stage replacement and input/output modification
- Validate plugin hook existence and function signatures during plugin loading via `inspect`
- Store imported plugin modules in `PluginConfig` with `invoke_replace`/`invoke_modify_input`/`invoke_modify_output` helpers
- Replace subprocess-based plugin execution with direct Python function invocation
- Update the `generate_phase_plan` and `generate_domain_context` stages to use the new API

### Standard hook signatures

```python
# Replace a stage entirely
def replace_<stage>(input_files: list[str], context: dict) -> list[str]: ...

# Modify stage input (workflow markdown) in-place
def modify_<stage>_input(input_file: str) -> None: ...

# Modify stage output files in-place
def modify_<stage>_output(output_files: list[str]) -> None: ...
```

### Files changed

| File | Change |
|------|--------|
| `src/plugin.py` | Complete rewrite: drop command API, add inspect validation, module loading, invoke helpers |
| `src/pipeline_setup.py` | Replace `run_plugin_command()` calls with `invoke_replace()`/`invoke_modify_output()` |
| `main.py` | Pass `plugin_config` to stage functions |

### DIFF
--- a/main.py
+++ b/main.py
@@ -188,6 +188,7 @@ def run_pipeline(
     one_phase=False,
     extra_call_edges_path=None,
     only_spec=False,
+    plugin_config=None,
 ):
     if not os.path.isdir(proj_dir):
         print(f"[Pipeline] ERROR: proj_dir does not exist or is not a directory: {proj_dir}")
@@ -227,10 +228,15 @@ def run_pipeline(
 
     # Stage 1: generate phase.json (input: target code → phases.json)
     # Stage 2: generate domain context (input: phases.json → domain context files)
+    phase_stage = plugin_config.get_stage("generate_phase_plan") if plugin_config else None
+    context_stage = plugin_config.get_stage("generate_domain_context") if plugin_config else None
+    plugin_root = plugin_config.root if plugin_config else None
+
     print("[Pipeline] Stage 1/6: Generating phase plan...")
     _run_generate_phases(
         proj_dir, work_dir, script_dir, resume=resume,
         submodules=submodules,
+        plugin_stage=phase_stage, plugin_root=plugin_root,
     )
 
     phases_modified = _post_process_phases(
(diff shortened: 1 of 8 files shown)


Xayahhh (approved):
good

Issue #129, opened 2026-07-15
[Feature Request] Add an interactive wizard for configuring OpenCode LLM providers


xuxian73:
## Problem

Configuring an LLM service for FM-Agent's OpenCode backend currently requires users to manually edit both:

1. The project-level `.env` file
2. The OpenCode configuration file, usually `~/.config/opencode/opencode.json`

According to `docs/config_llm.md`, users must manually keep the following values aligned:

- `LLM_API_KEY`
- `LLM_API_BASE_URL`
- `LLM_MODEL`
- `OPENCODE_MODEL_PROVIDER`
- The provider ID, adapter, endpoint, and model entry in `opencode.json`

This process is error-prone, especially for users who are unfamiliar with OpenCode's JSON format, use custom LLM endpoints, or already have an OpenCode configuration that must be preserved.

## Proposed solution

Provide an interactive configuration wizard for setting up the OpenCode LLM provider without requiring users to write JSON manually.

For example:

```bash
python3 scripts/configure_llm.py
```

The exact command and implementation location are open for discussion.

The wizard should prompt for:

1. Provider ID, such as `openrouter`, `deepseek`, or a custom value
2. Provider display name
3. API protocol:
   - OpenAI-compatible
   - Anthropic-compatible
4. API base URL
5. Model ID
6. API key
7. Whether to validate the generated configuration

It should then update both the project `.env` file and the appropriate OpenCode configuration file.

## Example interaction

```text
$ python3 scripts/configure_llm.py

FM-Agent LLM configuration

Provider ID: openrouter
Provider name [OpenRouter]: OpenRouter

API protocol:
  1. OpenAI-compatible
  2. Anthropic-compatible
Select [1]: 1

API base URL [https://openrouter.ai/api/v1]:
Model ID: anthropic/claude-sonnet-4.6
API key: ********

The following files will be updated:
  - /path/to/FM-Agent/.env
  - ~/.config/opencode/opencode.json

Continue? [Y/n]: y

✓ Updated .env
✓ Registered openrouter/anthropic/claude-sonnet-4.6 in OpenCode
✓ Configuration syntax is valid
```

## Expected output

The wizard should synchronize the FM-Agent environment variables:

```dotenv
LLM_API_KEY=<user-provided-key>
LLM_API_BASE_URL=https://openrouter.ai/api/v1
LLM_MODEL=anthropic/claude-sonnet-4.6
FM_AGENT_MODEL_BACKEND=opencode
[...]

PR #143, opened 2026-07-21
Add an interactive LLM setup wizard for the OpenCode backend

okcomputer2000:
## Summary

This PR adds an interactive configuration wizard for FM-Agent's OpenCode backend:

```bash
python3 scripts/configure_llm.py
```

The wizard collects the provider id, provider name, API protocol, base URL, model id, and API key, then safely updates the local configuration without requiring users to hand-edit OpenCode JSON.

## What it changes

- Adds `scripts/configure_llm.py`
- Adds the configuration logic in `src/configure_llm.py`
- Updates `README.md`, `README_zh.md`, and `docs/config_llm.md` to document the new workflow

The wizard now:

- updates `fm-agent.toml` with the selected non-secret LLM settings
- updates `.env` with `LLM_API_KEY` only
- removes common legacy LLM override keys from `.env` so they do not silently shadow `fm-agent.toml`
- detects the platform-appropriate OpenCode config path
- creates or merges the matching provider entry into `opencode.json`
- shows a preview before writing
- backs up existing files before modification
- writes files atomically
- refuses to overwrite invalid existing JSON or TOML
- keeps the API key out of `opencode.json` and references it via `{env:LLM_API_KEY}`

## Notes on current main

This is implemented against the current `main` branch architecture, not the older workflow described in the original issue text.

On current `main`, FM-Agent itself uses `fm-agent.toml` as the non-secret source of truth and injects the OpenCode provider configuration at runtime via `OPENCODE_CONFIG_CONTENT`. The wizard still synchronizes the user's local `opencode.json` so standalone OpenCode usage stays aligned, but FM-Agent runtime behavior continues to follow the current upstream design rather than reintroducing the old multi-file manual configuration model.

## Protocol support

The wizard supports both:

- OpenAI-compatible endpoints via `@ai-sdk/openai-compatible`
- Anthropic-compatible endpoints via `@ai-sdk/anthropic`

It validates that the selected adapter matches the chosen API protocol before writing.

## Conversion to issue requirements

This PR addresses issue #129 by providing a user-facing, safe configuration flow that:

- eliminates the need to manually write provider JSON
- updates both FM-Agent config and OpenCode config
- preserves unrelated OpenCode providers, models, plugins, and settings
- keeps secrets only in `.env`
- remains idempotent when rerun for the same provider/model pair


okcomputer2000:
`.env` backups are now written outside the repository worktree (under the system temp dir), so the wizard no longer leaves untracked secret-bearing backup files that can be accidentally committed.

haoran-ding (changes requested):
I do not think it is necessary to implement the wizard with 2 python files. Why not implement it in only one python file? Besides, running command should be `uv run python xxx` rather than `python xxx`.

okcomputer2000:
I merged the two wizard Python files into a single one – it’s simpler and cleaner now.
I updated the run command to use `uv run python xxx` (instead of plain `python xxx`) to ensure dependencies are properly managed.

The changes have been pushed.

okcomputer2000:
> The point about making LLM setup explicit through an interactive OpenCode backend wizard is worth separating from generic provider support. In real workloads, the hard part is usually not sending one request; it is keeping cost, latency, and response shape predictable once users switch providers.
> 
> I would have the wizard run separate checks for auth, model listing, streaming, tool support, and usage reporting. That gives users actionable failures instead of a generic bad-provider config.
> 
> I am working through similar issues while testing an OpenAI-compatible layer for official Chinese models, where normalized usage and fallback behavior matter a lot. Do you expect most FM-Agent users to configure local endpoints, hosted OpenAI-compatible APIs, or both?

Thanks, that makes sense. I agree the wizard should surface actionable failures rather than a generic “bad provider” result.

For this PR I’m keeping the scope focused on safe setup and config synchronization, but I can see value in adding optional checks for auth, model listing, streaming, and usage/tool support as a follow-up validation path.

FM-Agent is intended to stay provider-agnostic and support both local endpoints and hosted OpenAI-compatible APIs. I don’t have hard usage data to claim one is dominant, so I’d prefer to keep the wizard flexible for both.

okcomputer2000:
Addressed the configuration review feedback across the LLM wizard and runtime checks.

- Added a focused `uv run python src/configure_llm.py set ...` command for updating individual non-secret `[llm]` TOML settings, including model, provider, base URL, backend, effort, and API style. It previews changes, creates a backup, and warns when legacy dotenv or live shell variables would still take precedence.

- The interactive wizard now treats the active FM-Agent configuration file as the source of truth. It honors `FM_AGENT_CONFIG` from either the launching environment or the project `.env`, rather than always updating the repository-default `fm-agent.toml`.

- Selecting `auto`, `codex-cli`, or `claude-cli` now removes stale non-secret LLM dotenv overrides that could shadow the selected backend. Before removing them, the wizard migrates existing `LLM_MODEL` and `LLM_EFFORT` values, including dotenv `export ...` syntax, into the active TOML file. This preserves custom local CLI model and reasoning-effort choices instead of silently falling back to defaults.

- The migration preserves an intentionally empty model value as well, so local CLI execution continues to omit `--model` when that was the prior configuration. Existing effort values are also preserved verbatim so values accepted by the selected CLI are not discarded during migration.

- Added warnings for live shell overrides such as `LLM_API_KEY`, `LLM_MODEL`, `LLM_EFFORT`, `FM_AGENT_MODEL_BACKEND`, and OpenCode-related variables. The wizard reports the exact `unset ...` command needed because a child process cannot modify its parent shell environment.

- OpenCode configuration discovery now honors both override mechanisms:
  - `OPENCODE_CONFIG` remains the highest-priority explicit config-file path.
  - When only `OPENCODE_CONFIG_DIR` is set, the wizard writes to `<dir>/opencode.jsonc` when it exists, otherwise `<dir>/opencode.json`.
  This prevents the wizard from updating an inactive global config while a custom OpenCode config directory continues to override it.

- Updated the startup environment preflight to be backend-aware. For resolved local CLI backends, it skips OpenCode-specific API-key, `oh-my-openagent`, and comment-checker checks, while retaining the independent CodeGraph environment check.

- Updated `README.md`, `README_zh.md`, and `docs/config_llm.md` to document the local-backend migration behavior, override precedence, `OPENCODE_CONFIG_DIR` resolution, and backend-aware preflight behavior.

nhaorand:
@richardchen874-sys Thank you very much for the feedback! Please submit a new issue to describe your need and we will think about how to do that. This PR will focus on the current issue.

haoran-ding (approved):
Thanks

PR #146, opened 2026-07-22
Fix incremental selection of same-named methods

### LINKED ISSUE #122: Incremental mode over-verifies same-named functions in one file
In `--incremental` mode, the step that decides which changed functions to verify detects changes with un-qualified (bare) function names. When two classes in the same source file define a method with the same name, changing one method selects **both** for verification, so the untouched one is verified too.

## Reproduction

A single source file with two classes that share a method name, defined out-of-line:

```cpp
// storage.cpp
struct Buffer { int size = 0; };

class LocalStorage {
public:
    int Flush(Buffer &buf);
    void Reset(Buffer &buf);
};

class RemoteStorage {
public:
    int Flush(Buffer &buf);
    void Reset(Buffer &buf);
};

int LocalStorage::Flush(Buffer &buf) {
    int threshold = 1024;
    if (buf.size > threshold) return -1;
    return 0;
}
void LocalStorage::Reset(Buffer &buf) { buf.size = 0; }

int RemoteStorage::Flush(Buffer &buf) { return 0; }
void RemoteStorage::Reset(Buffer &buf) { buf.size = 0; }
```

1. Run the pipeline once as the incremental baseline.
2. Modify **only** `LocalStorage::Flush` (e.g. raise the threshold `1024` → `4096`) and commit.
3. Run the incremental pipeline against the baseline.

## Observed

Change detection is correct (`1 modified`), but mapping that one bare name back to the extracted files doubles it:

```
[Stage 5/10] ... -> 1 changed file(s): 0 added, 1 modified, 0 removed function(s).   # correct: only LocalStorage::Flush changed
[Stage 9/10] ... [specs] seeded 5 function(s) ... (2 changed, 5 relevant).           # bare "Flush" mapped to BOTH members
[Stage 10/10] ... [verify] running reasoner on 2 function(s)...                       # RemoteStorage::Flush verified too
```

The detector reports the change as the un-qualified name `Flush`; the un-qualified name matches both `LocalStorage::Flush` and `RemoteStorage::Flush`, so the untouched one is pulled into verification.

## Impact

- Incremental mode does redundant work: it re-verifies same-named methods that were not touched, defeating the point of scoping to the actual diff. The extra cost scales with how many classes in the file share a method name (common in C++, e.g. `Flush`/`Reset`/`Close`).
- The over-selection is name-collision-driven, so which methods get dragged in is not obvious from the diff.

Not a crash and nothing is dropped — it is a safe over-approximation (it never under-selects), only wasteful.

gesi519:
Fixes #122

## Summary

Incremental change detection previously retained bare function names from regex extraction. When multiple C++ classes in the same file defined a method with the same name, such as `LocalStorage::Flush` and `RemoteStorage::Flush`, the changed method could be mapped back to every matching extracted function.

This change maps current functions to CodeGraph's class-qualified identifiers when CodeGraph and regex extraction agree on the function spans. If they do not align, it safely falls back to the existing unqualified-name behavior.

## Changes

- Preserve qualified identifiers such as `LocalStorage::Flush` for added and modified functions.
- Keep the existing safe fallback when extraction results cannot be aligned.
- Add regression tests for same-named methods defined by different classes in one source file.


Dshuishui:
Nice, this cleanly fixes the demonstrated case. One coverage limitation worth noting:

The alignment gate compares regex vs CodeGraph **on the current file**, and the regex extractor doesn't pick up inline in-class member definitions (only free functions and out-of-line members). So a file that mixes an inline member with same-named out-of-line methods trips the count guard and the whole file falls back to unqualified names — reviving the exact over-verification this PR targets. Inline members are very common in C++, so this fallback path is easy to hit.

Minimal repro:

```cpp
struct Buffer { int size() const { return n; } int n = 0; };   // inline → CodeGraph sees it, regex doesn't
class LocalStorage  { public: int Flush(Buffer&); };
class RemoteStorage { public: int Flush(Buffer&); };
int LocalStorage::Flush(Buffer& b)  { /* changed */ }
int RemoteStorage::Flush(Buffer& b) { return 0; }
```

Changing only `LocalStorage::Flush`, `_collect_changed_functions` logs `could not align CodeGraph and regex functions (3 vs 2); using unqualified names` and reports `modified=['Flush']`, so both `Flush` files are selected again.

Since the mismatch comes from regex missing functions CodeGraph has, would it be worth deriving the base-revision functions from CodeGraph too (e.g. indexing the old commit in a temp checkout)? Then both sides would use class-qualified identities and the regex/CodeGraph divergence would go away entirely.

gesi519:
Implemented the suggested dual-CodeGraph comparison for CodeGraph-supported languages.

Incremental change detection now indexes the baseline commit in a temporary Git worktree and compares CodeGraph function maps for both revisions using qualified identifiers.

Erlang is intentionally out of scope for this change. It does not use the new dual-CodeGraph comparison and continues to follow its existing extraction path unchanged.

Added regression coverage for a file containing an inline `Buffer::size()` member plus same-named out-of-line `Flush` methods.

End-to-end validation now shows:
- `Buffer::size` is extracted by CodeGraph;
- 1 modified function is detected;
- the reasoner runs on 1 function only.

Dshuishui:
Thanks! This direction looks good — CodeGraph on both revisions, with the coverage gate addressing the P1.

It currently conflicts with main — could you rebase / resolve the conflict? Should be mergeable after that.

One non-blocking follow-up for a later PR (doesn't need to hold this one up): the coverage gate uses the regex extractor as the completeness floor, but neither extractor is guaranteed complete — regex misses C++ inline members and can over-report, and CodeGraph can mis-name or miss functions of its own. So any detection built on comparing extracted functions can still miss a real change whenever the extractor it trusts is wrong for that file.

A more robust direction worth considering later: anchor detection on git's line-level diff (which needs no extraction and is always reliable), and use CodeGraph/regex only to map the changed lines onto functions — if a changed line can't be attributed to any extracted function, re-check the whole file rather than dropping it. That way "which functions changed" is guaranteed by git, and the extractors only narrow the scope instead of being the thing that can miss a change. Not a blocker for this PR.

Dshuishui:
Thanks. This PR brings back some code that was already removed from main — it looks like part of what #116 deleted in `src/incremental_reasoner.py` came back along with the changes when they were moved onto the latest main. Could you take a look?

Dshuishui:
There's still an issue. The Files changed tab currently mixes in changes that others already merged into main — README.md, README_zh.md, main.py, `src/pipeline_setup.py`, the new `src/spec_generation_and_verification.py`, etc. (16 files, ~1400 lines) — because this branch merged main in and GitHub is still diffing against the base recorded when the PR was opened. That makes it hard to be sure nothing else was reverted along the way; the two spec helpers earlier slipped in exactly this way.

Dshuishui (approved):

PR #148, opened 2026-07-22
Revert artifact contract regression from PR #142

Xayahhh:
## Summary

- revert `a6f06a9` (`Enforce artifact contract for plugin stage pass/replace paths`)
- preserve `2e42996` and its plugin interface for phase-plan and domain-context customization
- restore `_phase_plan_schema_errors` and the pre-follow-up setup behavior

## Reason

The follow-up commit removed `_phase_plan_schema_errors` while leaving live call sites in the normal and resume phase-plan paths. Those paths fail at runtime with `NameError`. Reverting only the second commit restores the validator without removing the plugin interface introduced by the first commit.

## Verification

- confirmed the resulting tree matches `2e42996`
- `python3 -m py_compile main.py src/plugin.py src/pipeline_setup.py src/file_utils.py src/extract.py src/generate_topdown_layers.py src/incremental_reasoner.py src/entry_reasoning_pipeline.py`
- `git diff --check public/main...HEAD`

### DIFF
--- a/main.py
+++ b/main.py
@@ -15,7 +15,6 @@ from src.file_utils import (
     _json_file_is_valid,
     _get_incomplete_verification_files,
     _is_under_submodules,
-    load_phases,
 )
 from src.verification import streaming_reasoner
 from src.extract import run_extraction, EXT_TO_LANG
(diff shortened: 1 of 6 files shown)


haoran-ding (approved):
Approved

PR #157, opened 2026-07-26
Refactor extract_functions plugin handling and support incremental pipeline

### LINKED ISSUE #145: 插件接口设计 （Part 2.2）
Part 1 has been completed. See: https://github.com/fmagent-project/FM-Agent/issues/131

A plugin can replace or modify the input and output of each stage. In Part 1, this was implemented with command-line commands. Now it needs to be changed to Python functions.

0. Remove the required "replace_cmd" and "output_process" fields in Part 1 from plugin.json.
1. Define the parameters and return values for the replacement and modification functions of each stage.
    - Since the input and output of each stage are files, a stage replacement function should take one or more file paths as parameters and return one or more file paths.
    - Each stage should have two modification functions: one for the input file and one for the output file. Each function should take one file path as its parameter and return nothing.
2. Each plugin should include a Python file that implements these functions as needed.
3. When a plugin is loaded, the system should automatically check:
    - whether its plugin.json matches the implemented functions;
    - whether the function parameters and return values are defined correctly.

In this issue, you only need to make these changes for the third stage (Extracting functions from source file).

tanbing117:
Closes #145

> **Depends on:** #151 — review and merge that PR first.

### Motivation

The plugin system currently supports customization of several pipeline stages, but the `extract_functions` stage was only handled in the full pipeline. The incremental pipeline directly called `run_extraction()`, bypassing the plugin mechanism. This made plugin behavior inconsistent between full and incremental runs.

This change centralizes the `extract_functions` stage handling and enables the same plugin behavior in both pipelines.

### Changes

- Add a shared `_handle_extract_functions()` helper in `src/pipeline_setup.py`
  - Supports `none`, `pass`, `replace`, and `modify` plugin modes
  - Handles plugin output validation for replaced extraction results
  - Keeps extraction behavior consistent between full and incremental runs

- Update the full pipeline to use `_handle_extract_functions()` instead of maintaining its own plugin branching logic

- Update the incremental pipeline to use `_handle_extract_functions()` for function re-extraction

- Add runtime validation for plugin hook results:
  - `replace_*` functions must return `list[str]` with all elements being `str`
  - `modify_*` functions must return `None`

- Cache resolved plugin hook functions via `_resolve_hook()` to avoid repeated dynamic lookup

- Add `output_files` optional parameter to `run_extraction()` so `modify_output` hooks receive the full artifact list

- Add `extract_functions` to `_REPLACE_SPEC` and `_MODIFY_SPEC` for load-time signature validation

### Files changed

| File | Change |
|------|--------|
| `src/pipeline_setup.py` | +66: `_handle_extract_functions` handler |
| `src/plugin.py` | +42: runtime guards, hook cache, extract_functions specs |
| `src/extract.py` | +10: `output_files` param |
| `main.py` | -50 +2: replace inline dispatch with handler call |
| `src/incremental_reasoner.py` | -10 +4: replace direct call with handler + import cleanup |

### DIFF
--- a/main.py
+++ b/main.py
@@ -165,6 +165,7 @@ def run_pipeline(
         submodules=submodules,
         plugin_stage=phase_stage,
         plugin_root=plugin_root,
+        plugin_config=plugin_config,
     )
 
     phases_modified = _post_process_phases(
@@ -182,6 +183,7 @@ def run_pipeline(
         resume=resume and not phases_modified,
         plugin_stage=context_stage,
         plugin_root=plugin_root,
+        plugin_config=plugin_config,
     )
 
     # Build (or rebuild) the codegraph index if codegraph is installed. Both
--- a/main.py
+++ b/main.py
@@ -8,7 +8,7 @@ from src.file_utils import (
     _json_file_is_valid,
     _is_under_submodules,
 )
-from src.extract import run_extraction, EXT_TO_LANG
+from src.extract import EXT_TO_LANG
 from src.generate_topdown_layers import generate_topdown_layers
 from src.spec_generation_and_verification import run_spec_generation_and_verification
 from src.incremental_reasoner import run_incremental_pipeline
@@ -24,6 +24,7 @@ from src.pipeline_setup import (
     _run_generate_phases,
     _post_process_phases,
     _run_generate_domain_context,
+    _handle_extract_functions,
 )
 from src.domain_knowledge import (
     collect_domain_knowledge_paths,
@@ -197,7 +198,7 @@ def run_pipeline(
     # force=False on resume preserves already-specced extracted files; on a fresh
     # run fm_agent/ was just wiped so it is equivalent to force=True.
     print("[Pipeline] Stage 3/6: Extracting functions from source files...")
-    run_extraction(proj_dir, work_dir=work_dir, force=not resume, verbose=True)
+    _handle_extract_functions(proj_dir, work_dir, plugin_config, force=not resume)
 
     # Copy system_prompt.md to spec_prompts/system_prompt.md
     spec_prompts_dir = os.path.join(work_dir, "spec_prompts")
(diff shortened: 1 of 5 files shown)


haoran-ding (changes requested):
The PR changes many code irrelevant to the issue and may be incorrect, such as some code in pipeline_setup.py. Please carefully review the code you push.

haoran-ding:
By the way, another PR #150 also aims to close #145. Do you write the incorrect issue number?

tanbing117 (commented):
Thanks for the review!

You're right that this PR currently includes more changes than it should.

Part of the diff noise comes from the fact that this PR is stacked on top of #151, which hasn't been merged yet. As a result, GitHub is also showing changes from that PR, including unrelated parts of `pipeline_setup.py`.

For the Stage 3 plugin support itself, the main refactoring is introducing a shared `_handle_extract_functions()` helper. The intention is only to avoid duplicating the same pass/replace/modify branching logic in both the full and incremental pipelines.

I've removed the runtime guard and hook-cache changes from this PR so that it stays focused on the scope of this issue.

Thanks again for pointing this out. I'll update the PR accordingly.

tanbing117:
Thanks for the review!

You're right that this PR currently includes more changes than it should.

Part of the diff noise comes from the fact that this PR is stacked on top of #151, which hasn't been merged yet. As a result, GitHub is also showing changes from that PR, including unrelated parts of `pipeline_setup.py`.

For the Stage 3 plugin support itself, the main refactoring is introducing a shared `_handle_extract_functions()` helper. The intention is only to avoid duplicating the same pass/replace/modify branching logic in both the full and incremental pipelines.

I've removed the runtime guard and hook-cache changes from this PR so that it stays focused on the scope of this issue.

Thanks again for pointing this out. I'll update the PR accordingly.

PR #158, opened 2026-07-26
Fix incremental Erlang change detection

### LINKED ISSUE #156: Incremental mode never detects Erlang function changes
## Summary

In `--incremental` mode, change detection (`_collect_changed_functions`) extracts the old and new versions of each changed file with the regex extractor (`extract_functions_from_file`). That extractor returns `[]` for Erlang — `LANG_CONFIG["erlang"]["body"] == "external"`, i.e. Erlang is handled by the ELP backend, which the change-detection path never calls. So a changed `.erl` function produces no `added`/`modified`/`removed` entry, the file is dropped from the changed set, and every downstream stage (scope, spec update, verification) seeds from that empty set. The changed Erlang function is never re-analyzed or verified.

This is independent of ELP being installed: the change-detection path uses the regex extractor, not ELP, so the blind spot exists even when ELP indexes Erlang correctly for the full extraction and stale-file reconciliation.

## Reproduction

A minimal Erlang module, committed as the baseline, then a body change to `add/2`:

```erlang
-module(mymod).
-export([add/2, greet/1]).

add(A, B) ->
    A + B.          %% baseline  ->  A + B + 1  (changed)

greet(Name) ->
    io:format("hello ~s~n", [Name]).
```

Run the incremental pipeline against the baseline commit.

## Observed

The `.erl` file produces no detected change:

```
extract_functions_from_file('mymod.erl', 'erlang') -> []
_collect_changed_functions(proj, baseline)         -> {}      # add/2 changed, nothing detected
```

The identical body change in a C++ file **is** detected (`modified: ['add']`), so the empty result is Erlang-specific.

## Impact

Changes to Erlang functions are silently missed in incremental mode: the modified function is never re-specced or re-verified, so a bug introduced in a changed Erlang function is not caught. Full (non-incremental) runs are unaffected — they extract Erlang via ELP.

gesi519:
Fixes #156 
## Summary

Fix incremental change detection for Erlang functions.

## Root cause

The incremental path used the regex extractor, which intentionally returns no functions for Erlang because Erlang extraction is handled by ELP.

## Changes

- Parse current and baseline Erlang source through ELP document symbols.
- Compare ELP function IDs and source bodies to classify changes.
- Add a regression test for a modified Erlang function.

### DIFF
--- a/src/incremental_reasoner.py
+++ b/src/incremental_reasoner.py
@@ -54,7 +54,10 @@ from .opencode_trace import run_opencode_traced
 from .llm_client import _llm_provider_client, _llm_json_call, build_llm_cli_command
 from .scope import _parse_issue_signals, rank_functions_in_file
 from .languages.codegraph import CodeGraphExtractor, try_codegraph_init
-from .languages import erlang as erlang_language
+from .languages.registry import (
+    extract_incremental_sources,
+    supports_incremental_source_extraction,
+)
 from .verification import _verify_single_file, _validate_single_bug, _generate_validation_summary, EXT_TO_LANG as _VERIFY_EXT_TO_LANG
 from .domain_knowledge import (
     format_domain_knowledge_bullets,
@@ -358,14 +361,13 @@ def _collect_changed_functions(proj_dir, old_commit_id, submodules=None):
     version, then compared by source text.
 
     Returns a dict mapping each changed file's absolute path to a dict with keys "added",
-    "removed", and "modified", each a sorted list of function names. For every
-    non-Erlang language that CodeGraph can index, both revisions are compared using
-    its class-qualified identifiers. Erlang uses its ELP semantic backend, while
-    CodeGraph-unavailable non-Erlang files retain the previous regex comparison.
-    Files with no detectable function-level change are omitted. Raises
+    "removed", and "modified", each a sorted list of function names. Languages
+    with a registered semantic source backend use it for both revisions; other
+    languages use CodeGraph when available, then the regex fallback. Files with
+    no detectable function-level change are omitted. Raises
     subprocess.CalledProcessError if proj_dir is not a git repository or
-    old_commit_id is not a valid commit, and RuntimeError if ELP cannot extract
-    a changed Erlang revision.
+    old_commit_id is not a valid commit, and RuntimeError if a registered source
+    backend cannot extract a changed revision.
     """
     # Pathspecs limiting git to recognized source-file extensions (e.g. "*.py", "*.cpp").
     pathspecs = [f"*.{ext}" for ext in EXT_TO_LANG]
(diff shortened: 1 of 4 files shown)


haoran-ding (changes requested):
All code specific to a programming language should stay under languages/. The other files only invoke the unified interface for analyzing code. The other files should not have code like `lang_key != "erlang"`. Otherwise, it will break the modularity.

src/incremental_reasoner.py
```diff
@@ -447,23 +453,60 @@ def _funcs_from_commit(rel_path, lang_key, ext):
...
+                proj_dir, lang_key, current_source_batches[lang_key]
+            ) if current_source_batches[lang_key] else {}
+        )
+        baseline_result = (
+            extract_incremental_sources(
+                proj_dir, lang_key, baseline_source_batches[lang_key]
+            ) if baseline_source_batches[lang_key] else {}
+        )
+        if current_result is None or baseline_result is None:
+            raise RuntimeError(
```
Dshuishui: When the change set contains an Erlang file and `extract_incremental_sources()` returns `None`, this raise happens before the per-file comparison loop, so otherwise valid results for other languages in the same change set are discarded as well.

Reproduced with one changed `.cpp` and one changed `.erl` in the same commit and no ELP installed: `main` still reports the C++ change, while this branch raises `RuntimeError`. ELP is an opt-in install (`install.sh --with-erlang`), and every existing backend capability in `registry.py` degrades when its backend is unavailable — `batch_extract` / `call_edges` return an empty dict and `function_spans` returns `None` for the caller to fall back on.

src/languages/erlang.py
```diff
@@ -482,6 +482,77 @@ def _symbol_line_span(symbol_range: dict) -> tuple[int, int]:
...
+            first_path = paths[0]
+            client.initialize(first_path, normalized_sources[first_path])
+            for path in paths[1:]:
+                client.open_document(path, normalized_sources[path])
+
+            result = {}
+            for path in paths:
+                symbols = client.request(
+                    "textDocument/documentSymbol",
+                    {"textDocument": {"uri": Path(path).as_uri()}},
```
Dshuishui: This request uses `Path(path).as_uri()`, while `open_document()` registers the document with `Path(path).resolve()`. For a project accessed through a symlink the two identify different documents:

```text
didOpen:        file://.../real/proj/src/mymod.erl
documentSymbol: file://.../link/proj/src/mymod.erl
ERROR: -32603 file not found
```

The same project works through its real path. The pattern is inherited from existing code — `_analyze_project_uncached()` does the same — so Erlang extraction on `main` is already empty under a symlinked project root; worth fixing both at once.

Dshuishui (commented):
Verified against `2238763`: Erlang body changes, added/removed functions, and arity changes are all detected correctly, and standalone C++ change detection is unchanged. Two issues, noted inline.

src/languages/erlang.py
Dshuishui: Confirmed fixed — detection works through a symlinked project path now.

src/incremental_reasoner.py
Dshuishui: This is better structured than before — other languages are now fully computed, and the check sits before Stage 5b so existing artifacts aren't wiped. Nothing blocking from my side.

One note for later: in a mixed-language repo (a rebar3 project with `src/*.erl` plus `c_src/*.c` NIFs, for example), an unavailable ELP still ends the run with a `RuntimeError` even though the C change was already detected, so the user gets no C results. In the same environment, full mode skips Erlang with a warning and produces the C side normally.

Separately, when ELP is unavailable the Stage 5b reconciliation deletes the Erlang extracted functions along with their spec sidecars. That's pre-existing on `main` and not something to fix in this PR — filed as #166.

Dshuishui (approved):
Verified end to end — Erlang detection works and the C++ path is unaffected. LGTM.

haoran-ding (approved):

PR #162, opened 2026-07-27
Add pre-run scope and cost estimates

heiyan-2020:
## What changed

- add a no-LLM `main.py <proj> --estimate` preflight mode
- add a one-shot `dashboard.py <proj> --estimate` view and keep the estimate panel visible during live runs
- inventory included/excluded directories, included/excluded supported source files, approximate function count, and all six analysis stages
- persist summaries from successful runs and use completed history to estimate duration, LLM calls, token volume, and cost ranges
- generate `fm_agent/estimate.json` before the first LLM stage of every full run and retain `fm_agent/history.jsonl` across fresh runs
- document the workflow in English and Chinese

## Why

Previously FM-Agent did code understanding and extraction only after a run started, while the dashboard reported only usage that had already occurred. Users had no way to inspect scope or understand likely time and cost before committing to a run.

The new preflight performs only local source scanning/extraction. Historical ranges are scaled by function count (file count fallback), use only completed runs, and are explicitly labelled **ESTIMATE**. If no completed history exists, the UI reports that the range is unavailable instead of fabricating a value.

## Validation

- `uv run python -m unittest discover -s tests -v` — 9 tests passed
- `uv run python -m compileall -q main.py dashboard.py src tests`
- `uvx ruff check src/run_estimate.py tests/test_run_estimate.py --select E4,E7,E9,F,I,UP`
- `uvx ruff check main.py dashboard.py --select E9,F63,F7,F82`
- `git diff --check`
- manually exercised `uv run python main.py . --estimate`
- manually rendered `uv run python dashboard.py . --estimate`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -218,5 +218,6 @@ docs/superpowers/
 
 .opencode/
 .codegraph/
+fm_agent_estimate/
 CLAUDE.md
 AGENTS.md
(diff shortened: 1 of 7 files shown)


xuxian73:
Superseded by #163, recreated from the verified local commit under the intended xuxian73 account.

PR #164, opened 2026-07-27
fix(reasoner): ignore braces in multiline block comments

okcomputer2000:
Fix brace-depth tracking for brace-delimited languages when block comments span multiple lines.

The scanner now preserves block-comment state across lines, so `{` and `}` inside `/* ... */` comments no longer affect nesting depth. After `*/`, code later on the same line is still scanned normally.

This prevents `_split_into_blocks_braced` from using incorrect syntactic boundaries for C, C++, Java, JavaScript/TypeScript, Rust, and similar languages.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,6 +11,8 @@ dependencies = [
     "litellm>=1.50.0",
     "socksio>=1.0.0",
     "pydantic-settings>=2.2",
+    "tree-sitter>=0.25,<0.27",
+    "tree-sitter-cpp>=0.23.4,<0.24",
 ]
 
 [project.scripts]
--- a/src/languages/cpp.py
+++ b/src/languages/cpp.py
@@ -1,4 +1,88 @@
+import re
+
 from src.languages.codegraph import CodeGraphExtractor
+
+
+_LINE_PREFIX = re.compile(r"^Line \d+: ?")
+
+
+def _line_end(node) -> int:
+    """Return the inclusive source line containing ``node``'s end."""
+    row, column = node.end_point
+    return row - 1 if column == 0 and row > node.start_point[0] else row
+
+
+def _function_block_nodes(root):
+    """Return safe top-level C++ block boundaries for one function."""
+    stack = list(reversed(root.named_children))
+    while stack:
+        node = stack.pop()
+        if node.type == "function_definition":
+            body = node.child_by_field_name("body")
+            if body is not None and body.type == "compound_statement":
+                return [child for child in body.named_children if child.type != "comment"]
+            if body is not None and body.type == "try_statement":
+                return [body]
+            function_try = next(
+                (child for child in node.named_children if child.type == "try_statement"),
+                None,
+            )
+            if function_try is not None:
+                return [function_try]
+            return None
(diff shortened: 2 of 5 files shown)


okcomputer2000:
I reverted the follow-up template-literal and regex-literal scanner changes, so the effective implementation is now back to the focused multiline block-comment fix from `4997795`.

One scope question remains: the current shared scanner still does not distinguish JavaScript/TypeScript template-literal text or regex literals from comment delimiters. A multiline template literal containing `/*` without a later `*/`, or a regex such as `/[/*]/`, can therefore still produce an incorrect brace depth.

This may indicate that a single shared brace scanner is reaching its limits. C/C++, Java, Rust, and JavaScript/TypeScript have materially different lexical rules, especially around raw strings, template literals, nested comments, and regex-versus-division ambiguity.

Should the follow-up work introduce language-specific brace-depth scanners, or at least separate the JavaScript/TypeScript lexer path from the C-style scanner? I would prefer to confirm that design direction before expanding this PR beyond the original multiline block-comment fix.

haoran-ding (changes requested):
The current code depends on regex to split code into blocks. Check if there are any third party lib that can do that. For example, the code now uses codegraph to construct call graph. Does codegraph help split the code into blocks?

okcomputer2000:
I checked the pipeline and the available options.

The current splitter is a heuristic used for LLM reasoning chunks, not for constructing compiler CFG basic blocks. It splits long functions so FM-Agent can generate a post-condition for each chunk and use it as the next chunk's pre-condition. Therefore, we still need splitting, but each split must occur only between complete syntax structures.

CodeGraph helps earlier in the pipeline: it indexes the project, extracts complete function spans, and builds the call graph. It uses Tree-sitter internally, in [`src/extraction/tree-sitter.ts`](https://github.com/fmagent-project/codegraph/blob/v1.3.0-fmagent.1/src/extraction/tree-sitter.ts#L18), but the data it persists and FM-Agent currently reads is only a symbol/call graph. It does not expose statement-level AST ranges, block nodes, or CFG edges, so the current CodeGraph integration cannot directly split a function into reasoning blocks.

Tree-sitter is the most suitable third-party option. We can keep CodeGraph for function boundaries and call edges, then use Tree-sitter on each extracted function to obtain complete top-level statement/control-structure ranges and group them near `GRANULARITY`. This avoids splitting inside comments, strings, `if`/loop bodies, C++ raw strings, or JS/TS literals.

With the current implementation, a split point is selected when the character scanner believes the brace depth has returned to the function-body depth after approximately `GRANULARITY` lines. It can therefore be confused by language-specific lexical constructs and may choose a line that is not a real statement boundary. With Tree-sitter, split points would instead come from parsed source ranges: for example, complete expression statements, declarations, `if`/`switch` statements, loops, and `try` blocks. The resulting chunks would remain close to the existing size target, but no chunk boundary would be placed inside one of those syntax structures.

okcomputer2000:
What do I need to do next?

haoran-ding:
You can invoke treesitter. If the programming language is not supported by treesitter, fall back to regex. But write the code in a modularity style.

PR #165, opened 2026-07-28
feat(plugin): pluginize all pipeline stages and add entry reasoning plugin

QiaoyiZheng:
## Summary

This PR extends FM-Agent's plugin system from the Stage 3-only implementation to all six stages of the full pipeline, and reimplements entry-point-scoped reasoning through the new multi-stage plugin interfaces.

### Multi-stage plugin configuration

- Allow one `plugin.json` to configure multiple pipeline stages.
- Validate each stage independently according to its supported modes, fields, and Python function signatures.
- Add an optional plugin-level `configure_function(options: dict) -> None` hook for receiving runtime options.
- Make `plugin.py` optional for pass-only and Markdown-only plugins.

### Pass, replace, and modify modes

All six stages now support the applicable plugin modes:

1. `generate_phase_plan`
2. `generate_domain_context`
3. `extract_functions`
4. `collect_file_list`
5. `generate_topdown_layers`
6. `generate_specs_and_verification`

The plugin contracts support:

- `pass`: consume an existing valid stage output.
- `replace`: replace the built-in stage with a Python function.
- `modify`: modify the stage input and/or canonical output while retaining the built-in implementation.
- `replace_md`: replace the built-in workflow Markdown for Stages 1, 2, and 6.
- `modify_md`: append plugin instructions to the built-in workflow Markdown for Stages 1, 2, and 6.

Plugin outputs are validated before downstream stages consume them.

### Stage 4 and Stage 5 data flow

Stage 5 now consumes the function list produced by Stage 4 instead of rescanning every extracted function from `phases.json`.

This makes `fm_agent_file_list.json` the authoritative function selection result and prevents functions filtered out by a Stage 4 plugin from being reintroduced during top-down layer generation.

### Entry-reasoning plugin

Add the bundled `entry_reasoning` plugin, which implements entry-point-scoped analysis through the standard full pipeline:

- Stage 1 selects the source files needed for the reachable call paths.
- Stage 3 continues to use FM-Agent's built-in function extraction.
- Stage 4 selects the extracted functions that participate in the analysis.
- Stages 5 and 6 consume the Stage 4 selection normally.

Example:

```bash
uv run python main.py /path/to/project \
  --plugin entry_reasoning \
  --entry-func "main-py::application_entry"
```

Optional terminal functions can restrict selection to paths between the entry and specified endpoints:

```bash
uv run python main.py /path/to/project \
  --plugin entry_reasoning \
  --entry-func "main-py::application_entry" \
  --end-func "services::statistics-py::calculate_total"
```

The entry-reasoning plugin supports the direct full pipeline and may be combined with resume, only-spec, one-phase, domain knowledge, supplemental call-graph edges, and custom bug validation. It is not compatible with incremental, isolate, or submodule mode.

### Documentation

- Rewrite the English and Chinese plugin-development documentation.
- Document all six stage contracts and exact Python signatures.
- Document multi-stage JSON configuration and workflow Markdown hooks.
- Add entry-reasoning usage and compatibility notes to both READMEs.

## Validation

The implementation was validated with:

- Unit tests for plugin loading, configuration, and all six stage hooks.
- Pass, replace, and modify behavior.
- Valid and invalid replacement artifacts.
- Input and output modification hooks.
- Workflow `replace_md` and `modify_md` behavior.
- Full real-pipeline consumption for all six stages.
- Resume behavior.
- Entry-reasoning selection on multi-file, multi-function projects.
- Entry-to-terminal function selection.
- Stage 4 to Stage 5 function filtering.

Local test plugins and test fixtures used for real-pipeline validation are intentionally excluded from the commit.
```


haoran-ding (changes requested):
Why do you need `validate_topdown_layer_files`? I think it is useless because generating topdown layers is deterministic without invoking LLMs.

QiaoyiZheng:
> Why do you need `validate_topdown_layer_files`? I think it is useless because generating topdown layers is deterministic without invoking LLMs.

generate_topdown_layers() itself is deterministic and does not invoke an LLM. However, validate_topdown_layer_files() primarily validates the plugin boundary rather than the built-in algorithm. Stage 5 pass/replace/modify hooks and the Stage 6 input hook can provide or mutate top-down JSON independently of the built-in generator. The validator enforces path containment, schema consistency, phase identity, uniqueness, and exact correspondence with the authoritative Stage 4 file list before downstream consumption. The validation immediately after the unmodified built-in path is defensive and could be made conditional, but the validator itself is required for plugin-provided artifacts.

please review

QiaoyiZheng:
Add backward compatibility for the legacy CLI arguments. Please review.

Issue #166, opened 2026-07-28
Incremental mode deletes Erlang extracted functions and their specs when ELP is unavailable


Dshuishui:
## Summary

In `--incremental` mode, Stage 5b reconciles the extracted-functions tree against what each source file currently contains, so that functions removed from the source do not leave orphaned specs behind. For Erlang, that list comes from the ELP backend. When ELP is unavailable, the backend yields nothing — and an empty result is indistinguishable from "this file genuinely has no functions", so every extracted file for that source is treated as orphaned and deleted, together with its `.spec.json` / `.info.json` sidecars. Only a warning is logged and the run reports success.

Reconciliation runs over every source file in `phases.json`, not just the changed ones, so unchanged `.erl` files are affected as well.

## Reproduction

1. With ELP available, run extraction on a small Erlang project so `fm_agent/extracted_functions/<file>-erl/` holds its functions, and let spec generation write the `.spec.json` / `.info.json` sidecars.
2. Make ELP unavailable (e.g. drop it from `PATH`).
3. Run incremental mode against a baseline commit — including a run whose change set contains no `.erl` file.

## Observed

```
WARNING: ELP Erlang analysis unavailable for <proj>: [Errno 2] No such file or directory: 'elp'

before reconciliation: ['mymod__add__2.erl', 'mymod__add__2.info.json', 'mymod__add__2.spec.json',
                        'mymod__greet__1.erl', 'mymod__greet__1.info.json', 'mymod__greet__1.spec.json']
after  reconciliation: []
```

The source file is unchanged and still defines both functions.

## Impact

The sidecars are LLM-generated artifacts from a previous run, so a transient backend failure silently discards work that has to be paid for again.

Triggers are not limited to a missing install: an ELP session that times out (180 s default), an OTP/ELP version mismatch, a project root without `rebar.config`, or a CI image without the optional `--with-erlang` dependencies all produce the same empty result.

## Expected

Reconciliation should skip source files whose semantic backend could not be consulted, and surface the degraded coverage, instead of reading an unavailable backend as an empty file.

## Notes

Pre-existing on `main`. PR #158 adds a guard that covers part of this — when the change set contains an Erlang file, the run stops before Stage 5b and the artifacts survive — but a run that touches no `.erl` file still reaches reconciliation and deletes them.

PR #170, opened 2026-07-31
fix: Incremental mode deletes Erlang extracted functions and their specs when ELP is unavailable

lztmdrclht:
#### Problem:
 [#166](https://github.com/fmagent-project/FM-Agent/issues/166)
Stage 5b Proofreading: Treat empty results returned by ELP (due to unavailability) as “no functions in the file,” and treat all .erl, .spec.json, and .info.json generated files for that source file as orphaned files and delete them.

#### Solution:
Have the Erlang backend pass through backend failures in `function_spans`; have `_function_spans` return the `backend_available` flag; and have the reconciliation process skip the step and explicitly report degraded coverage based on this.

#### Files Affected
src/languages/base.py (New): Define the `BackendUnavailableError` exception to avoid circular imports.
src/languages/registry.py: Export this exception and update the LanguageHandler documentation.
src/languages/erlang.py: Add `_analysis_or_raise`; modify `function_spans` to raise an exception when ELP fails, while `batch_extract` and `call_edges` retain their original fault-tolerant behavior.
src/extract.py: _function_spans now returns (spans, raw_lines, backend_available); catches backend failures; no longer falls back to regex indent extraction for Erlang external body type errors.
src/incremental_reasoner.py: _reconcile_extracted_dir skips and returns False when backend_available=False; _remove_stale_extracted collects skipped files; run_incremental_pipeline outputs a degraded coverage warning.
src/scope.py, src/entry_reasoning_pipeline.py: Unpacking has been updated to return a three-element tuple.


src/languages/erlang.py
```diff
@@ -693,20 +694,40 @@ def _analysis_or_none(proj_dir: str) -> ErlangAnalysis | None:
...
+        raise
+    except Exception as exc:
+        logging.warning("ELP Erlang analysis unavailable for %s: %s", proj_dir, exc)
+        raise BackendUnavailableError(
+            f"ELP Erlang analysis unavailable for {proj_dir}: {exc}"
+        ) from exc
 
+def batch_extract(proj_dir: str) -> dict:
+    """Return ``{abs_filepath: [(function_id, body)]}`` for Erlang files."""
+    return _analysis_or_empty(proj_dir).functions
```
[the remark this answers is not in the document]
lztmdrclht: #158 already fix this

lztmdrclht:
@haoran-ding This PR needs to be reviewed.

src/languages/registry.py
```diff
@@ -11,6 +11,7 @@
 from src.languages import typescript as _typescript
 from src.languages import erlang as _erlang
 
+from src.languages.base import BackendUnavailableError
```
Dshuishui: The `LanguageHandler` docstring just below (line 22, and the "returns None so the caller can fall back to the regex extractor" note on line 30) still describes `function_spans` as returning `None` when the backend is unavailable, but Erlang now raises `BackendUnavailableError`. Worth documenting the new contract here so whoever adds the next language knows whether they should raise it.

Also, since this re-export currently has no users (see the note on `extract.py`), an `__all__` entry would keep linters from flagging it as an unused import.

src/extract.py
```diff
@@ -8,6 +8,7 @@
 from src.file_utils import is_file_ready, _is_test_file
 from src.languages.codegraph import canonicalize
 from src.languages.registry import batch_extract_all, function_spans_for_file
+from src.languages.base import BackendUnavailableError
```
Dshuishui: This imports the exception directly from `src.languages.base`, bypassing registry as the single entry point for language capabilities — every other module under `src/` reaches language backends through `registry`. Importing it from `registry` here instead would keep that consistent, and would give the re-export added there a user.

src/verification.py
```diff
@@ -525,4 +525,4 @@ def _generate_validation_summary(proj_dir):
         json.dump(summary, f, indent=2, ensure_ascii=False)
     os.replace(tmp_path, summary_path)
     logging.info(f"Validation summary written to {summary_path}")
```
Dshuishui, line 527: This file's only change is dropping the trailing newline; the content is unchanged. Suggest restoring it.

Dshuishui (commented):
Verified — nothing major. A few small suggestions inline.

src/languages/registry.py
lztmdrclht: It is fixed.

src/verification.py
Dshuishui: The other two look good. This one went the other way though — there's now a whitespace-only line and the trailing newline is still missing (`...errors}")\n    `), so `git diff --check` reports both `trailing whitespace` and `new blank line at EOF`. All that's needed is the newline at the end of the last line, no spaces. Since this file is unrelated to the change, just restoring it to the main version works.

lztmdrclht: It is fixed.

Dshuishui (approved):

PR #173, opened 2026-08-02
feat(plugin): redesign all pipeline stages as Python plugins

### LINKED ISSUE #160: Add a new plugin for entry reasoning
Refactor entry_reasoning_pipeline as a plugin

QiaoyiZheng:
## Summary

- Add a unified Python plugin contract for all six FM-Agent pipeline stages.
- Support `pass`, `replace`, and `modify` modes.
- Standardize every plugin hook as:
  `def hook(proj_dir: str) -> None`.
- Preserve existing inline stage implementations with minimal hook boundaries.
- Implement Entry Reasoning as a standard plugin.
- Support explicit and implicit Entry plugin activation.
- Support `--end-func` call-chain termination.
- Validate only the selected plugin during normal runs.
- Preserve full plugin scanning for `--list-plugin`.
- Add English and Chinese plugin documentation.

## Entry Reasoning

Entry selection now runs through standard Stage 1 and Stage 4 hooks:

- Stage 1 computes entry reachability and narrows the phase plan.
- Stage 4 keeps only selected extracted functions.
- Stage 5 and Stage 6 operate on the filtered function set.

The old dedicated Entry pipeline has been removed.

## Fixes from real pipeline validation

- Preserve source extensions in sidecar names:
  `foo.py.spec.json` and `foo.py.info.json`.
- Propagate Entry filtering into topdown generation and verification.
- Avoid loading unrelated plugins when one plugin is selected.
- Integrate the latest public `main`, including #170.

Closes #160


haoran-ding (changes requested):
I think you may also need to remove some  code in entry reasoning about plugin.

haoran-ding:
Sorry, it seems okay to not touch entry reasoning now.

QiaoyiZheng:
please review

haoran-ding (approved):

PR #177, opened 2026-08-03
Fix/pr168 sidecar pair reverify

### LINKED ISSUE #167: spec generation: LLM drops source extension when writing sidecar files, causing false "missing spec" retries and pipeline exit
## Summary

During spec generation, the LLM occasionally writes `.spec.json` / `.info.json` sidecar files **without the source file extension** (e.g. `Preprocessor::preprocess_source.spec.json` instead of `Preprocessor::preprocess_source.rs.spec.json`). Because `is_file_ready()` expects the full filename including extension (`{file_path}.spec.json`), the pipeline falsely concludes these functions have no specs, retries 5 times, and then exits with an error. The repeated retries also burn significant API quota and memory.

## Reproduction

1. Run FM-Agent on a Rust project:

   ```bash
   uv run python main.py /path/to/dsv4-ccc --submodule src/frontend
   ```

2. Wait for spec generation on files with `::` in the name (e.g. `Preprocessor::preprocess_source.rs`).

3. Observe that some functions get `xxx.spec.json` instead of `xxx.rs.spec.json`.

4. Pipeline retries 5 times, then exits:

   ```
   [Pipeline] ERROR: Stage 6 Layer 2 failed after 5 attempts. 
   2 function(s) are still missing valid specs.
   ```

## Root Cause

The batch prompt tells the LLM:

```
For each function file `<function-file>`, write TWO JSON files in the SAME directory:

`<function-file>.spec.json`:
```

For `Preprocessor::preprocess_source.rs`, the LLM interprets `<function-file>` as the basename without extension, producing `Preprocessor::preprocess_source.spec.json` instead of `Preprocessor::preprocess_source.rs.spec.json`.

This is inconsistent — for some functions the LLM writes the correct name (`ExprId::of.rs.spec.json`), for others it drops `.rs`.

huangwei021230:
## Summary

  Closes #167 and builds on #168 by fixing the remaining timing, retry, and re-
  verification issues around extension-dropped spec sidecars.

  Some spec-generation runs write `foo.spec.json` / `foo.info.json` instead of the
  expected `foo.py.spec.json` / `foo.py.info.json`, causing false missing-spec retries
  and preventing verification from completing.

  ## Changes

  - Clarify in all spec-generation prompts that sidecar names must retain the source
  extension.
  - Normalize only complete, schema-valid `.spec.json` / `.info.json` pairs.
  - Reject ambiguous mappings instead of assigning sidecars to the wrong source file.
  - Normalize each batch as soon as its producer finishes, allowing
  `streaming_reasoner` to verify it while other batches are still running.
  - Rescan once when all producers finish to close the final producer/watcher race.
  - Invalidate stale verification and bug-validation artifacts when sidecars change.
  - Clear alternate sidecars before retries to prevent mixing files from different
  generation attempts.
  - Preserve normal retry behavior for incomplete or invalid sidecar pairs.

  This keeps normalization and verification correctly ordered without introducing a
  layer-wide synchronization barrier.


haoran-ding (changes requested):
It is not reasonable to directly raise AmbiguousSidecarError. If that error happens, the code should delete illegal files and re-generate sidecars rather than directly crash.

PR #188, opened 2026-08-07
feat(entry): support multiple entry functions

### LINKED ISSUE #184: Support multiple --entry-func values
## Motivation

Entry-scoped analysis currently accepts only one `--entry-func`, while some targets expose multiple entry points that should be analyzed together in one run.

## Proposed interface

Allow `--entry-func` to accept multiple FQNs:

```bash
uv run python main.py /path/to/project \
  --entry-func module_a::entry1 module_b::entry2 \
  --end-func module_c::end1 module_d::end2
```

Without `--end-func`, retain the union of the call graphs reachable from all entries. With end functions, retain the union of all valid entry-to-end paths.

A single `--entry-func` value should remain backward compatible.

yeunmer2006:
Closes #184

## Summary

- Allow `--entry-func` to accept one or more space-separated function FQNs.
- Normalize programmatic string/list inputs and de-duplicate entries while preserving their first-seen order.
- Build one multi-source BFS call graph containing the union reachable from all requested entries.
- When `--end-func` is provided, retain the union of valid entry-to-end chains and treat each end function as terminal.
- Validate every requested entry and report all missing entry FQNs together.
- Exempt every entry source file from test-file filtering, while forcing only source files that survive end pruning into `phases.json`.

## Usage

Keep the project positional before the variadic function options:

```bash
uv run python main.py <proj_dir> \
  --entry-func main-py::application_entry \
  --end-func services/statistics-py::calculate_total
```

Multiple entries and ends are accepted in the same invocation:

```bash
uv run python main.py <proj_dir> \
  --entry-func main-py::entry_a api-py::entry_b \
  --end-func services-py::end_a services-py::end_b
```
Without `--end-func`, FM-Agent analyzes the union reachable from all entries. With end functions, it analyzes only functions lying on at least one valid requested entry-to-end path.

### DIFF
--- a/README.md
+++ b/README.md
@@ -216,8 +216,6 @@ uv run python main.py <proj_dir> [--resume] [--all-bugs] [--domain-knowledge FIL
 | `--bug-validator FILE`       | Use a custom Markdown prompt for bug validation instead of the built-in `md/bug_validator.md`. |
 | `--isolate`                 | Run against an isolated git worktree snapshot of the project instead of the project directory itself. |
 | `--submodule PATH [PATH ...]` | Only process source code under one or more subdirectories of `proj_dir`. |
-| `--entry-func FQN [FQN ...]` | Run the dedicated entry-point pipeline from one or more entry functions. |
-| `--end-func FQN [FQN ...]` | With `--entry-func`, retain paths that reach one or more end functions and stop analysis at those functions. |
 | `--extra-edge FILE`         | Add supplemental caller-to-callee edges to the static call graph from a JSON file or directory. |
 | `--only-spec`               | Only generate behavioral specs; skip the reasoning and bug validation stages. Cannot be combined with `--incremental`. |
 | `--plugin NAME`             | Load a pipeline plugin from `plugins/NAME/`. |
@@ -235,6 +233,8 @@ def hook(proj_dir: str) -> None:
 
 See [Pipeline Plugins](docs/plugins.md) for the plugin layout, JSON
 configuration, execution modes, lifecycle, and trust boundary.
+
+`proj_dir` must be a git repository.
 
 To provide project-specific domain knowledge without editing FM-Agent's built-in prompts, pass one or more Markdown files:
 
--- a/README.md
+++ b/README.md
@@ -236,48 +236,6 @@ def hook(proj_dir: str) -> None:
 See [Pipeline Plugins](docs/plugins.md) for the plugin layout, JSON
 configuration, execution modes, lifecycle, and trust boundary.
 
-### Entry-point analysis
-
-Place `proj_dir` before `--entry-func`. A single entry function remains
-supported:
-
-```bash
-uv run python main.py <proj_dir> \
-  --entry-func main-py::application_entry
-```
-
-Pass multiple FQNs after one `--entry-func` to analyze several entry points in
-one run:
-
-```bash
-uv run python main.py <proj_dir> \
-  --entry-func main-py::entry_a api-py::entry_b
-```
-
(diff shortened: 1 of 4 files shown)


src/entry_reasoning_pipeline.py
```diff
@@ -505,13 +518,22 @@ def _run_entry_pipeline_inner(
...
+        # cannot reach any requested end_func may have been pruned along with its
+        # source file, so do not reintroduce that nonexistent path into the plan.
+        entry_source_files = dict.fromkeys(
+            _entry_func_source_rel(entry) for entry in entry_funcs
+        )
+        required_source_files = [
+            source_file
+            for source_file in entry_source_files
+            if source_file in keep_by_source
+        ]
```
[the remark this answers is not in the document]
yeunmer2006: Thanks, the stale workspace behavior is real, but it predates this change.
  On the base commit, a single-entry run resumed with a narrower --end-func scope
  already reuses the broader phases.json and extracted workspace in the same way.

  The required_source_files change here is additive: it prevents a pruned entry
  source from being forced back into a fresh plan, but does not introduce or
  define cross-scope resume behavior. Fully reconciling a changed resume scope
  would also require invalidating the phase plan, extracted functions, generated
  context/layers, specs, results, and CodeGraph state.

  Since --resume is intended to continue an interrupted run with the same entry
  scope, I suggest handling scope-change detection and workspace invalidation in
  a separate issue.

src/entry_reasoning_pipeline.py
```diff
@@ -430,23 +441,25 @@ def _select_functions_by_source(proj_dir, entry_func, end_funcs, extra_call_edge
     finally:
         shutil.rmtree(sel_dir, ignore_errors=True)
 
-    # Keep only functions on a call chain from entry_func to one of end_funcs.
+    # Keep only functions on a call chain from an entry to one of end_funcs.
     if end_funcs:
         unreachable = sorted(set(end_funcs) - set(call_graph))
-        call_graph = _restrict_to_chains(call_graph, entry_func, end_funcs)
+        call_graph = _restrict_to_chains(call_graph, entry_funcs[0], end_funcs)
```
lar0129: `_restrict_to_chains()` only performs reverse reachability from `end_funcs`; entry reachability has already been established by the preceding multi-source BFS. The `entry_func` parameter is unused, so passing `entry_funcs[0]` is misleading. Please remove the parameter and call `_restrict_to_chains(call_graph, end_funcs)` instead.

README.md
```diff
@@ -234,6 +236,46 @@ def hook(proj_dir: str) -> None:
 See [Pipeline Plugins](docs/plugins.md) for the plugin layout, JSON
```
lar0129, line 234: This dedicated section is more detailed than necessary. The new entries in the option table are sufficient; please keep those and remove this section and the corresponding section from `README_zh.md`.

main.py
```diff
@@ -549,8 +549,9 @@ def run_pipeline(
     parser.add_argument(
         "--entry-func",
         metavar="PATH",
+        nargs="+",
```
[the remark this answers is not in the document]
lar0129: This does not need to be solved, the position of proj_dir is fixed at the front

src/entry_reasoning_pipeline.py
yeunmer2006: Addressed. I removed the unused `entry_func` parameter from `_restrict_to_chains()`, updated its docstring, and simplified the call site to `_restrict_to_chains(call_graph, end_funcs)`.

haoran-ding (changes requested):
Please revert the change of readme.

haoran-ding (approved):
Thanks

PR #193, opened 2026-08-09
feat: Add a user-facing report index with search and filtering

lztmdrclht:
Closed, superseded by [#203](https://github.com/fmagent-project/FM-Agent/issues/203)

#### Objective [#179 ](https://github.com/fmagent-project/FM-Agent/issues/179):
After running FM-Agent, generate a `report.html` file to search for, view, and browse reports.

- Each row/entry displays the title, status, file, function, and code location.
- The search function supports text matching in titles, files, and functions.
- Users can combine search with filters.
- Users can sort the visible result set without losing their current filter settings.
- Technical details for each entry can be expanded or collapsed.
- The index is regenerated deterministically based on existing runtime artifacts, without invoking a large language model (LLM).
- The documentation explains how to locate and open the results index.

#### Features to Be Implemented
It is unclear what the intended behavior is when opening a code file within an HTML file, so this will be addressed later.


haoran-ding (changes requested):
Do not push unrelated files. Besides, You just need to generate html files to show bug found. So you just need to add one new python file to convert bug reports to html rather than changing so many files.

PR #194, opened 2026-08-10
fix(config): validate TOML files and base URLs

okcomputer2000:
## Summary

Fix LLM configuration handling for empty TOML files, quoted TOML keys, and invalid base URL ports.

## Changes

- Allow existing empty `fm-agent.toml` files to be initialized.
- Update bare, basic quoted, and literal quoted TOML keys without creating duplicate entries.
- Validate URL hostnames and ports before saving configuration.
- Preserve existing formatting, comments, backups, and atomic writes.

### DIFF
--- a/src/configure_llm.py
+++ b/src/configure_llm.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import argparse
 import json
 import hashlib
+import ipaddress
 import os
 import re
 import shutil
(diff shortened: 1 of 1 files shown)


okcomputer2000:
## Description

This PR fixes three configuration-handling issues in `src/configure_llm.py`.

The implementation keeps the existing format-preserving TOML editor and changes only the validation and key-matching logic required for these cases.

### Modification approach

- Distinguish an absent TOML file from an existing empty TOML file.
  - The write entry points now check `Path.is_file()` before reading the file.
  - An existing empty file is passed to the normal TOML update flow and initialized with an `[llm]` table.
  - A missing file is still rejected to avoid guessing or creating an unrelated project configuration.

- Recognize all supported TOML key forms.
  - The key matcher now accepts:
    - bare keys, such as `name`;
    - basic quoted keys, such as `"name"`;
    - literal quoted keys, such as `'name'`.
  - Matched key tokens are parsed with `tomllib` to obtain their semantic key name.
  - The original key spelling, quoting style, indentation, spacing, and trailing comments are preserved.
  - The editor tracks semantic keys, so quoted and unquoted representations update one entry instead of appending a duplicate key.
  - Basic quoted key escapes, such as `"na\u006de"`, are interpreted correctly.

- Validate URL hostnames before saving configuration.
  - `validate_base_url()` continues to require an absolute `http` or `https` URL.
  - It now checks that `parsed.hostname` is present and syntactically valid.
  - IPv4 and IPv6 addresses are validated with Python's standard `ipaddress` module.
  - DNS hostnames are checked by label, including label length, allowed characters, and hyphen placement.
  - IDNA hostnames and portless HTTP(S) URLs remain supported.

- Validate URL ports at the configuration boundary.
  - The implementation explicitly accesses `parsed.port`, which makes Python's URL parser reject non-numeric, negative, and out-of-range ports.
  - Those parsing errors are converted into `ConfigWizardError` before TOML or OpenCode configuration is written.

The current behavior treats `http://example.com:` as a portless URL because `urllib.parse` returns `None` for that port value. If the project decides that an explicitly empty port must be rejected, a small additional check can reject authorities ending in `:`.

heiyan-2020 (approved):

PR #197, opened 2026-08-11
fix(codegraph): keep FM-Agent's work directory out of the project index

### LINKED ISSUE #195: codegraph indexes FM-Agent's own artifacts as project source
## Summary

FM-Agent writes its artifacts to `<proj>/fm_agent/`, which sits inside the tree
codegraph indexes. Every extracted function therefore appears in the index twice
(the original source plus the extracted copy), and the FM-Agent scripts staged
under `spec_prompts/` are indexed as if they were project code.

**This happens in every mode, starting with the first run.**

## Why

Artifacts enter the index at two points:

1. **When the index is built** — the previous run's artifacts are still sitting in
   `fm_agent/`, so `codegraph init` picks them up while scanning the project
   directory. Guaranteed from the second run onwards.
2. **After the index is built** — oh-my-openagent runs `codegraph sync` at the start
   of every OpenCode session, scanning the project and pulling in whatever is on
   disk by then. FM-Agent starts a session per LLM stage, so the first one after
   Stage 3 syncs the freshly written artifacts in. (The MCP file watcher is not
   what does this: with `CODEGRAPH_NO_WATCH=1` the pollution is unchanged.)

The second point is why the very first run is affected, and why building the index
before writing artifacts does not avoid it.

## Observed

A Python project with four functions, plain full run, sampling the index throughout:

```
[t= 77.3s] nodes=6   edges=9      src/app.py                    <- just built, clean
[t= 79.3s] nodes=77  edges=131                                  <- two seconds later
           fm_agent/extracted_functions/src/app-py/*.py   (x4)
           fm_agent/spec_prompts/file_utils.py
           fm_agent/spec_prompts/generate_batch_prompts.py
           src/app.py
```

Function nodes by file — 43 of the 47 are FM-Agent's own:

```
 4  src/app.py                                     <- project code
 4  fm_agent/extracted_functions/src/app-py/*.py   <- extracted copies
22  fm_agent/spec_prompts/file_utils.py            <- FM-Agent's own source
17  fm_agent/spec_prompts/generate_batch_prompts.py
```

## Impact

FM-Agent's own output is unaffected; the cost falls on the index itself — which is
a shared asset in the user's project:

- Any symbol lookup in the project returns two hits, plus FM-Agent's internal
  functions
- The extra copies scale with the number of functions in the project
- Every artifact write makes the daemon re-parse and re-sync

Dshuishui:
Fixes #195.

FM-Agent writes its artifacts to `<proj>/fm_agent/`, inside the tree codegraph
indexes. Every extracted function ends up in the index a second time next to the
source it was copied from, and the scripts staged under `spec_prompts/` are
indexed as if they were project code — on a 4-function Python project that is 43
of 47 function nodes.

Ordering the run around it does not help, because artifacts get in two ways:
`codegraph init` picks up what a previous run left behind, and the codegraph MCP
daemon's file watcher syncs them in as they are written, seconds after the index
was built. `codegraph.json` is codegraph's own project-scoped exclude and both
paths honour it — its loader runs per index, per sync and per watch event.

## What this does

Before running `codegraph init`, append `"fm_agent/"` to `exclude` in the
project's `codegraph.json`, creating the file when absent.

- **Idempotent** — the entry is added once; later runs read the file and return.
- **Non-destructive** — other keys and existing `exclude` entries are kept. An
  unreadable file, non-object JSON, or a non-array `exclude` is left untouched
  with a warning rather than rewritten, matching how codegraph itself treats a
  config it cannot parse.
- **Announced** — the one run that writes it prints a line saying what changed
  and why; this is the only file FM-Agent adds outside its own work directory.
- **Skipped when codegraph is absent**, so no stray config is left behind.

The call sits in `try_codegraph_init`, the single point all five call sites go
through, so full, `--resume`, `--incremental`, `--entry-func` and `--isolate`
runs are all covered.

## Verification

Two full pipeline runs on a fresh Python fixture, sampling the index throughout:

| | index during the run |
|---|---|
| before | `nodes=6` → `nodes=77` two seconds after Stage 3 |
| after | `nodes=6` for the whole run |

Layering, extraction, specs and verification are unchanged (4 functions, 4
layers, `main → run_all → Engine::start → Engine::warmup`).

Also checked directly: an existing config keeps its other keys and gains only the
new entry; a config with comments or a non-array `exclude` is left byte-for-byte
untouched with a warning; a second call writes nothing and prints nothing.

## Not in scope

`try_codegraph_init` removes `.codegraph/` before rebuilding, without restoring
it if the rebuild does not happen. That is a separate defect with its own fix
(`codegraph index` instead of `rmtree` + `init`) and will be filed on its own.

### DIFF
--- a/src/languages/codegraph.py
+++ b/src/languages/codegraph.py
@@ -9,6 +9,7 @@ REGISTRY in src/languages/registry.py. No other files need to change.
 """
 
 import hashlib
+import json
 import logging
 import os
 import re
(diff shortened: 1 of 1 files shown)


Dshuishui:
Solved in the pinned build instead. `v1.5.0-fmagent.2` — the version #206 moves the pin to — carries `fm_agent` in codegraph's default ignore list, so there is no longer a reason to write a `codegraph.json` into a repository FM-Agent does not own.

Verified with a full run in each mode (plain, `--incremental`, `--entry-func`) on projects carrying neither a `codegraph.json` nor a `.gitignore`: no fm_agent node reaches the index.

Issue #204, opened 2026-08-11
Unify call edge representation across language backends


tanbing117:
##Background

FM-Agent currently supports multiple language backends (codegraph, regex-based extractors, and other language handlers). However, the call edge representation returned by these backends is inconsistent.

The current interface mainly uses:

{
    caller_fqn: {
        callee_fqn,
        ...
    }
}

This representation is sufficient for top-down layer generation, but it loses important edge metadata and makes it difficult for future analyses (e.g., security plugins) to consume call graph information consistently.

##Problem

The current call edge interface has several limitations:

1. Edge metadata is lost

Different backends may distinguish different edge types, such as:

normal function calls
constructor calls

However, the current FQN-set representation merges them into the same relation:

caller -> {callee1, callee2}

The edge type information cannot be preserved.

2. Call-site information is unavailable

The current representation only stores:

caller function -> callee function

It does not preserve where the call happens.

This prevents downstream analyses from reliably associating source-level information with a specific call site.

For example:

check(password)
check(token)

Both become:

foo -> check

and the two call sites cannot be distinguished.

3. Backend formats are inconsistent

Different language handlers currently return different formats:

dictionary-based edges:
{
    caller: {callee}
}
list-based edge objects:
[
    {
        "caller": "...",
        "callee": "..."
    }
]
custom edge objects

Every consumer must understand each backend's format separately.

This makes extending the call graph interface difficult.

##Proposal

Introduce a normalized call edge representation inside the language registry layer.

All language backends should be converted into:

[
    {
        "caller": FunctionId,
        "callee": FunctionId,
        "kind": "call" | "constructor",
        "language": str,
        "span": {
            "file": str,
            "start_line": int,
            "start_column": int,
        }
    }
]

The normalization layer should:

accept existing backend formats
validate required fields
normalize optional metadata
deduplicate edges at call-site granularity
preserve backend-specific information when available翻译

PR #205, opened 2026-08-11
Normalize call-graph edge representation across language backends

tanbing117:
Fixes#204

## Motivation

FM-Agent is integrating security analysis plugins (IFC, taint, crypto, resource) that perform cross-function data-flow analysis and need a precise call graph — distinguishing same-named functions, preserving edge kind, and carrying call-site location. The current `get_call_edges()` output (FQN-set dict) cannot provide these.

### src/languages/codegraph.py
- `get_call_edges()` returns a list of edge dicts `[{caller, callee, kind, span}]` instead of `{caller: {callee}}`.
- Edge kind (`calls` vs `constructor`) preserved.
- Each edge carries call-site location `{file, start_line, start_column}`.
- Results `ORDER BY` source location so call-site order aligns with source appearance.
- Dedup key includes file/line/col so multiple call sites of the same callee are all preserved.

### src/languages/registry.py
- `normalize_call_edges()` unifies dict (incl. single-string callee), list, and custom-object edge formats into one schema.
- Schema validation skips malformed edges with a warning (shared infrastructure, never silent).
- Dedup at call-site granularity (idempotent); includes language so C/C++ backends don't merge away.
- Span field-name aliases normalized to `{file, start_line, start_column}`.
- Empty language overwritten by caller language; custom objects extracted via getattr (`__slots__`/`@property` compatible).

### src/generate_topdown_layers.py
- Adapts the main pipeline consumer to the new list format by converting back to the legacy `{caller: {callee}}` shape at the boundary, preserving existing behavior exactly.

## Design

codegraph → list{caller, callee, kind, span}
              → registry.normalize_call_edges()
              → main pipeline  and  → ProgramIndex for plugins

This is the first step toward a shared ProgramIndex contract; edge kind and call-site metadata are preserved end-to-end for future plugin consumption.

### DIFF
--- a/src/generate_topdown_layers.py
+++ b/src/generate_topdown_layers.py
@@ -315,9 +315,14 @@ def _build_call_graph(phase_files, proj_dir, global_stem_to_fqns=None, extra_cal
     phase_langs = {_detect_lang_from_ext(fp) for fp, _ in phase_files if _detect_lang_from_ext(fp)}
     registry_edges_list, registry_langs = call_edges_all(proj_dir, phase_langs)
     # codegraph edges are [{"caller": fqn, "callee": fqn, "kind": k}, ...]
-    registry_edges: dict = {}
+    # normalize_call_edges guarantees caller/callee are present, but use .get()
+    # defensively in case an edge bypasses the normalization layer.
+    registry_edges = defaultdict(set)
     for edge in registry_edges_list:
-        registry_edges.setdefault(edge["caller"], set()).add(edge["callee"])
+        caller = edge.get("caller")
+        callee = edge.get("callee")
+        if caller and callee:
+            registry_edges[caller].add(callee)
 
     for filepath, module_name in phase_files:
         fqn = fqn_map[filepath]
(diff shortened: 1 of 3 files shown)


Dshuishui:
Thanks for taking this on! Just two small things that might be worth addressing:

1. Some of the comments in `normalize_call_edges` are in Chinese — might be worth switching to English to match the rest of the codebase.

2. The `kind` values do not seem fully unified yet: codegraph emits `"calls"` while the dict backends and #204's schema use `"call"`. A consumer filtering on `kind == "call"` would miss every codegraph edge — mapping it in the normalization layer would probably close this.

One note: this is producer-side only for now — `_build_call_graph` collapses the edges back into `{caller: {callee}}`, so `span` and `kind` are not consumed anywhere yet. Worth keeping in mind that the consumer side needs to pick these up before the change has any visible effect.

tanbing117:
> 感谢你承担这个任务！有两个小问题值得讨论：
> 
> 1. 里面有些注释是中文的——可能值得切换成英文，以匹配代码库的其他部分。`normalize_call_edges`
> 2. 这些值似乎还没有完全统一：码图输出，dict 后端和`kind``"calls"`[统一跨语言后端的呼叫边缘表示 #204](https://github.com/fmagent-project/FM-Agent/issues/204)的模式使用 。消费者在 上过滤时会错过所有码图边缘——在归一化层映射它可能会关闭这个过程。`"call"``kind == "call"`
> 
> 需要说明的是：目前这只是生产者端的——将边缘收缩回 ，所以 和 还没有被消耗。值得记住的是，消费者方面需要先了解这些信息，才能让变更产生明显影响。`_build_call_graph``{caller: {callee}}``span``kind`

@Dshuishui Thank you for the review！ I have made the corresponding modifications. Regarding the note，I'm waiting for this PR to be merged, and then I will integrate the security plugins into FM-Agent's plugin framework uniformly, so I will use these changes in the follow-up PRs.

Dshuishui (approved):

Issue #211, opened 2026-08-17
ArkTS is extracted by regex although codegraph supports it


Dshuishui:
REGISTRY in `src/languages/registry.py` has no `arkts` entry, so `.ets` files never reach codegraph and fall back to the regex extractor. The reason given in `src/languages/codegraph.py` is:

```python
# ArkTS is omitted (not supported by codegraph).
```

codegraph does support ArkTS, and the build we currently pin already carries it — no codegraph upgrade needed.

PR #214, opened 2026-08-18
Route ArkTS .ets extraction through CodeGraph (#211)

okcomputer2000:
Register ArkTS in FM-Agent’s language registry and CodeGraph language mapping so .ets files use the pinned CodeGraph backend instead of the regex extractor. Add the ArkTS adapter for function extraction, function spans, and call edges, including TypeScript-compatible constructor edge handling while preserving the existing fallback behavior when CodeGraph is unavailable.

### DIFF
--- a/src/languages/arkts.py
+++ b/src/languages/arkts.py
@@ -24,16 +24,24 @@ def remove_comments(code: str) -> str | None:
     source_start = 0
     source_end = len(source)
     if tree.root_node.has_error:
-        prefix = b"class __FM_AGENT_COMMENT_WRAPPER__ {\n"
-        source = prefix + source + b"\n}"
-        try:
-            tree = parser.parse(source)
-        except (TypeError, UnicodeError, ValueError):
+        tree = None
+        for prefix in (
+            b"class __FM_AGENT_COMMENT_WRAPPER__ {\n",
+            b"struct __FM_AGENT_COMMENT_WRAPPER__ {\n",
+        ):
+            wrapped_source = prefix + source + b"\n}"
+            try:
+                candidate = parser.parse(wrapped_source)
+            except (TypeError, UnicodeError, ValueError):
+                return None
+            if not candidate.root_node.has_error:
+                source = wrapped_source
+                tree = candidate
+                source_start = len(prefix)
+                source_end += source_start
+                break
+        if tree is None:
             return None
-        if tree.root_node.has_error:
-            return None
-        source_start = len(prefix)
-        source_end += source_start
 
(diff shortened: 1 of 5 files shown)


Dshuishui (approved):

Dshuishui (commented):
Thanks for taking this on.

One problem worth flagging, pre-existing rather than introduced here: `arkts` supplies no `remove_comments`, so `.ets` sources fall back to the scanner in `parser._remove_func_comments`, which doesn't treat backticks as string delimiters. ``const url = `https://${host}/api`;`` reaches the reasoner as ``const url = `https:``.

okcomputer2000:
Implemented. ArkTS now uses its language-specific comment remover, so comment-like text inside backtick template literals is preserved instead of falling back to the legacy scanner.

src/languages/arkts.py
```diff
@@ -0,0 +1,67 @@
...
+        source = code.encode("utf-8")
+        parser = Parser(Language(ts_arkts.language()))
+        tree = parser.parse(source)
+    except (TypeError, UnicodeError, ValueError):
+        return None
+
+    source_start = 0
+    source_end = len(source)
+    if tree.root_node.has_error:
+        prefix = b"class __FM_AGENT_COMMENT_WRAPPER__ {\n"
```
Dshuishui: This fallback only retries the extracted fragment inside a regular `class`. ArkUI declarative bodies such as `build()` need a `struct` context, so they still fail to parse here, `remove_comments()` returns `None`, and the caller drops back to the legacy scanner.

Extraction writes one function per file, so an isolated `build()` is exactly what the reasoner receives. In a full pipeline run with this commit installed, this:

```ts
Text(`https://${this.host}/api/list?page=${this.page}`)
```

still reaches the reasoner as:

```ts
Text(`https:
```

One direction worth trying: retry with a `struct` wrapper when the `class` wrapper fails. That parsed every ArkUI form I tried and left ordinary class methods unaffected, but my sample was small — worth verifying more broadly.

okcomputer2000: Thanks for catching this. I updated the ArkTS comment remover to retry isolated fragments in a `struct` wrapper after the existing `class` wrapper fails.

This preserves the original class-method path while allowing extracted ArkUI `build()` bodies to parse with the ArkTS grammar, including declarative component syntax such as `Column() { ... }`, `Text(...)`, and `ForEach(...)`. Once the struct-wrapped fragment parses successfully, comments are removed from Tree-sitter comment nodes while template literals and other string content remain unchanged.

PR #215, opened 2026-08-18
Resolve #212: Migrate Erlang analysis from ELP to CodeGraph

### LINKED ISSUE #212: Erlang is extracted through ELP although codegraph supports it
Erlang is one of the few languages that do not go through codegraph — it goes through ELP instead.

That costs us twice. First an external dependency: the user has to install the `elp` binary themselves, and without it the language has no analysis capability at all. Second, a fair amount of code exists only to serve it — an LSP client, its cache and fingerprinting, the diagnostic file it persists (where #202 came from), and a branch in incremental mode that exists solely for Erlang, splitting languages into a source-backend group and a codegraph group.

codegraph does support Erlang, and the build we currently pin already carries it — no codegraph upgrade needed.

okcomputer2000:
Resolves #212 by replacing the ELP-based Erlang backend with the pinned CodeGraph backend. Erlang function extraction, function spans, and call-graph edges now use CodeGraph through the existing language registry; the ELP LSP client, cache, diagnostics artifact, configuration, installer option, and Erlang-only incremental extraction path have been removed. Documentation now reflects that Erlang no longer requires a separately installed ELP toolchain.


haoran-ding:
We will still use ELP, which is more powerful.

PR #219, opened 2026-08-22
Fix Erlang call-edge FQNs for top-down layering (#218)

### LINKED ISSUE #218: Erlang call edges use an identity format the pipeline does not match, leaving the call graph empty
The call graph for Erlang projects is empty. `generate_topdown_layers` receives no Erlang call edges, so every function collapses into a single layer. Each function is then reasoned about in isolation, without its callees' specs as context — which is exactly what the top-down layering exists to provide.

ELP itself resolves the calls correctly; the loss happens after that.

### Reproduction

Two modules with a three-level call chain (`run -> greet -> format_msg`) plus one cross-module call:

```erlang
%% src/mymod.erl
-module(mymod).
-export([greet/1, run/0]).

greet(Name) ->
    Msg = format_msg(Name),
    io:format("~s~n", [Msg]),
    ok.

format_msg(Name) ->
    lists:concat(["hello ", Name]).

run() ->
    greet("world"),
    helper:aux(1).
```

```erlang
%% src/helper.erl
-module(helper).
-export([aux/1]).

aux(N) -> N * 2.
```

Run extraction and layer generation with ELP available. Extraction is fine:

```
WRITE: fm_agent/extracted_functions/src/mymod-erl/mymod__run__0.erl
WRITE: fm_agent/extracted_functions/src/mymod-erl/mymod__greet__1.erl
WRITE: fm_agent/extracted_functions/src/mymod-erl/mymod__format_msg__1.erl
WRITE: fm_agent/extracted_functions/src/helper-erl/helper__aux__1.erl
Validation passed: every extracted file contains exactly one function.
```

The call graph is not:

```
[TopdownLayers] Phase 2: 3 functions, 1 layers

 layer 0  src::mymod-erl::mymod__format_msg__1   callees=[]  callers=[]
 layer 0  src::mymod-erl::mymod__greet__1        callees=[]  callers=[]
 layer 0  src::mymod-erl::mymod__run__0          callees=[]  callers=[]
```

Three layers are expected.

### Suspected cause

Judging from the output, `call_edges` returns edges shaped like this:

```python
{('mymod__run__0', 'mymod-erl'): {'mymod__greet__1', 'helper__aux__1'}}
```

The caller is a tuple and the callee is a bare name, while downstream lookups are keyed by string FQNs such as `src::mymod-erl::mymod__run__0`. Our guess is that this identity mismatch is where the edges are lost — every other language backend was unified to the list-of-dicts form with string FQNs in #204, and Erlang appears to be the only one still on the older shape.

We have not dug further; this is only a rough direction.

### Environment

Reproduced on current `main` (`08c5edb`) with ELP 1.1.0 (`build-2026-08-10`) installed and working — all four functions extract successfully, so this is not an ELP availability problem.

okcomputer2000:
Fixes #218 by converting ELP call-hierarchy edges to the FQN format used by extracted Erlang functions and the top-down layer builder. Internal cross-module calls now reach the call graph, while external or invalid targets are excluded.

### DIFF
--- a/src/languages/erlang.py
+++ b/src/languages/erlang.py
@@ -493,7 +493,7 @@ def _persist_analysis(proj_dir: str, fingerprint: tuple, analysis: ErlangAnalysi
         for callee in sorted(callees)
     ]
     document = {
-        "schema_version": 1,
+        "schema_version": 2,
         "status": "success",
         "backend": "elp",
         "server_info": analysis.server_info,
--- a/src/languages/erlang.py
+++ b/src/languages/erlang.py
@@ -40,7 +40,7 @@ _PROJECT_CONFIG_FILES = ("elp.toml", "rebar.config", "rebar.lock")
 @dataclass
 class ErlangAnalysis:
     functions: dict[str, list[tuple[str, str]]]
-    edges: dict[tuple[str, str], set[str]]
+    edges: dict[str, set[str]]
     spans: dict[str, list[tuple[str, int, int]]] = field(default_factory=dict)
     server_info: dict | None = None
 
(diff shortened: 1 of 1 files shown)


okcomputer2000:
## Description

Fixes #218 by converting ELP call-hierarchy results into the same FQN format used for extracted Erlang functions and top-down layer generation.

ELP returns each outgoing target as a `CallHierarchyItem` with a canonical file URI. The backend now resolves that URI to a canonical path only for lookup, then maps it back to the logical `.erl` source path that FM-Agent indexed and extracted. FQNs are built from that logical path, preserving the extracted-function layout:

```text
<source-directory>::<source-file>-erl::<function-id>
```

This allows both same-module and cross-module Erlang calls to reach `call_edges_all()` and `generate_topdown_layers()` as standard `{"caller", "callee", "kind": "call"}` edges. Targets that do not correspond to an indexed Erlang source file, such as OTP or third-party library calls, are excluded.

The mapping also handles path aliases:

- A project root accessed through a symbolic link still matches ELP’s canonical file URIs.
- A source file that is itself a symbolic link keeps its logical extraction path in the FQN.
- Multiple logical source paths pointing to one physical Erlang file each retain their extracted-function identity and receive the resolved call edge.

The persisted `fm_agent/erlang_callgraph.json` diagnostic now records FQN endpoints and function FQNs. Its `schema_version` is bumped from `1` to `2` because these fields are structurally incompatible with the previous module-qualified/bare-name representation.

Dshuishui (approved):

PR #221, opened 2026-08-22
fix(spec): disambiguate qualified callee expectation matching

### LINKED ISSUE #213: Layer prompts can attach a caller's expectations to the wrong same-named callee
When a caller calls two functions whose bare names are identical, the `## CALLEE EXPECTATIONS FROM CALLERS` section of a layer-N prompt (N ≥ 1) can carry the expectations written for the other one.

`_callee_match_names` in `src/generate_batch_prompts.py` builds two candidates per callee — the full FQN and the bare name:

```python
names = [callee_fqn, callee_fqn.split("::")[-1]]
```

The FQN candidate never matches. FQNs look like `dir::base-ext::Class::method`, while the names an LLM writes into `.info.json` are class-qualified but carry no path prefix (`Class::method`), and `_info_line_mentions_name` tests `name in first_line` — the longer string searched inside the shorter one.

Matching therefore always falls through to the bare name, which every same-named callee satisfies, and the loop returns the first one it reaches.

A minimal case — two classes each with a `Flush`, both called from one function:

```cpp
int LocalStorage::Flush(Buffer &buf);   // returns -1 when buf.size > 4096
int RemoteStorage::Flush(Buffer &buf);  // always returns 0

int RunCheckpoint(Buffer &buf) {
    int rc = local.Flush(buf);
    if (rc != 0) return rc;
    return remote.Flush(buf);
}
```

The caller's `.info.json` records correct and distinct expectations for the two, but both sections of the generated prompt come out identical:

```
### What callers expect from ...::RemoteStorage::Flush:
int LocalStorage::Flush(Buffer& buf)
  Post-condition: Returns 0 when buf.size <= 4096; returns -1 when buf.size > 4096
```

`RemoteStorage::Flush` then generates its spec against a contract it does not satisfy, and the one actually written for it — including call-site context only the caller knows, such as "this call is only made after the local flush has already succeeded" — is lost.

The class name is present throughout; matching just tries a direction that cannot succeed and then falls back to a bare name that drops it. Testing whether the `.info.json` name is a suffix of the FQN would tell them apart.

yeunmer2006:
## Summary

- Prioritize exact FQN and qualified-suffix matches when resolving caller expectations.
- Support both C++-style `Class::method` and dot-qualified `Class.method` names.
- Match explicit `callee.info_names` aliases before attempting bare-name fallback.
- Prevent qualified entries from being incorrectly selected through bare-name matching.
- Preserve literal dots inside FM-Agent FQN components such as `foo.test-py`.

Closes #213

### DIFF
--- a/src/generate_batch_prompts.py
+++ b/src/generate_batch_prompts.py
@@ -2,6 +2,7 @@
 
 import argparse
 import json
+import logging
 import re
 from pathlib import Path
 from typing import Dict, List, Optional, Sequence, Tuple
@@ -56,6 +57,12 @@ COMMENT_PREFIX_BY_LANG = {
 }
 
 _QUALIFIED_NAME_SEPARATOR_RE = re.compile(r"::|\.")
+_ERLANG_NAME_RE = re.compile(
+    r"^(?P<module>[A-Za-z_][\w@]*)\s*:\s*"
+    r"(?P<function>[A-Za-z_][\w@]*)/(?P<arity>\d+)$"
+)
+
+logger = logging.getLogger(__name__)
 
 
 def parse_args() -> argparse.Namespace:
@@ -138,6 +145,11 @@ def extract_callee_spec_from_info(
     names = _callee_match_names(callee_fqn, aliases or ())
     callees = info_dict.get("callees", [])
     if not isinstance(callees, list):
+        logger.warning(
+            "No callee expectation matched FQN %s (aliases=%s, entries=invalid)",
+            callee_fqn,
+            list(aliases or ()),
+        )
         return None
 
     named_callees = []
@@ -148,13 +160,14 @@ def extract_callee_spec_from_info(
         if not isinstance(name, str):
             continue
         normalized_name = _name_without_call_decoration(name)
-        named_callees.append((callee, name, normalized_name))
+        bare_name = _bare_name_for_matching(normalized_name)
+        named_callees.append((callee, name, normalized_name, bare_name))
 
-    for callee, name, _normalized_name in named_callees:
+    for callee, name, _normalized_name, _bare_name in named_callees:
         if _info_name_matches_fqn(name, callee_fqn):
             return callee
 
-    for callee, name, _normalized_name in named_callees:
+    for callee, name, _normalized_name, _bare_name in named_callees:
         if any(
             _info_name_matches_alias(name, alias)
             for alias in aliases or ()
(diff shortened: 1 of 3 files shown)


src/generate_batch_prompts.py
```diff
@@ -155,11 +180,81 @@ def _callee_match_names(callee_fqn: str, aliases: Sequence[str]) -> List[str]:
...
+    """Split C++- and dot-qualified source names into components."""
+    return [part for part in _QUALIFIED_NAME_SEPARATOR_RE.split(name) if part]
+
+
+def _is_qualified_name(name: str) -> bool:
+    """Return whether a source-level name contains a known qualifier."""
+    return bool(_QUALIFIED_NAME_SEPARATOR_RE.search(name))
+
+
+def _name_without_call_decoration(name: str) -> str:
```
Dshuishui: The trailing paren group is stripped unconditionally. That's correct for `Engine::start(int n)`, but wrong when the parentheses hold the function name:

```
FQN          src::store-ets::Counter::constructor
.info entry  'Counter (constructor)'

08cd3153     matched 'Counter (constructor)'
184f671e     None
```

Stripping leaves `Counter`, which no longer aligns with an FQN ending in `::constructor`. This entry is actual `.info.json` output from an ArkTS project run.

Dshuishui (dismissed):
The suffix-alignment approach is right — same-named callees no longer bleed into each other. But several spellings went from matching to returning nothing, so I'd hold the merge until those are addressed.

**1. Entries are dropped even when there is no ambiguity**

In each row below the `.info.json` holds only that one entry and the project has no same-named function. All of them matched correctly before this change:

| `.info.json` entry | FQN | before | after |
|---|---|---|---|
| `Vec<T>::push_back` | `...::Vec::push_back` | matched | **None** |
| `(*Store).Flush` | `...::Store::Flush` | matched | **None** |
| `ns::Engine::start` | `...::Engine::start` | matched | **None** |
| `engine.start` | `...::Engine::start` | matched | **None** |
| `Counter (constructor)` | `...::Counter::constructor` | matched | **None** |

The first two rows are the case raised in https://github.com/fmagent-project/FM-Agent/pull/221#discussion_r3849037204, and stripping `<T>` and `(*)` before the comparison resolves them. The last three are not reachable that way: `ns` is a segment the FQN never had, `(constructor)` is lost by the paren-stripping step, and `engine` has no correspondence to `Engine`.

**2. The cause is the `continue` on line 167**

It decides whether to allow the bare-name fallback based on **whether the name contains `::` or `.`**. But whether an entry can be confused with another depends on **whether the caller calls two functions with the same bare name** — and those cover different sets: the first spans every qualified name, the second only occurs on a collision. None of the five rows above involves a collision, yet all are blocked.

**3. Erlang's name format isn't recognized**

`_QUALIFIED_NAME_SEPARATOR_RE` only knows `::` and `.`, while Erlang writes `mod:fun/arity`, so those read as bare names:

```
FQN            src::mymod-erl::mymod::greet
.info entries  ['helper:greet/1', 'mymod:greet/1']
before         'helper:greet/1'
after          'helper:greet/1'      <- should be mymod:greet/1
```

#213 still reproduces there.

**4. Nothing is reported when an entry can't be found**

When any of the above happens, the whole caller-expectations section is simply absent from the prompt, with no trace in the logs. A log line when a callee's expectations are discarded would make this class of problem visible in real runs.

Dshuishui (approved):

